PR #4, opened 2019-10-17
Enable all managers

hopkiw:
Enable all managers appropriate to each OS
Correct a few small bugs

### DIFF
--- a/google_guest_agent/main.go
+++ b/google_guest_agent/main.go
@@ -115,10 +115,10 @@ func runUpdate() {
 	}
 
 	var wg sync.WaitGroup
-	mgrs := []manager{newWsfcManager(), &addressMgr{}}
+	mgrs := []manager{&addressMgr{}}
 	switch runtime.GOOS {
 	case "windows":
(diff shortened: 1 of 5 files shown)


google_guest_agent/main.go
```diff
@@ -115,11 +115,12 @@ func runUpdate() {
 	}
 
 	var wg sync.WaitGroup
-	var mgrs []manager
-	if runtime.GOOS == "windows" {
-		mgrs = []manager{newWsfcManager(), &addressMgr{}, &winAccountsMgr{}}
-	} else {
-		mgrs = []manager{&clockskewMgr{}}
+	mgrs := []manager{newWsfcManager(), &addressMgr{}}
```
adjackura: Do we want to enable wsfcManager?

hopkiw: i assumed so since it was before.....

adjackura: I meant for Linux

hopkiw: oh, duh. no we don't, moved into the windows block.

PR #6, opened 2019-10-21
Add basic Windows instance setup for metadata route, fix issue with Windows addresses

adjackura:
Updated logging to pass in stdout and serial port logging
Added basic instanceSetup function, currently just has a setting to add the route to the metadata server on the first interface on Windows
Added a few common functions for controlling forwarding rules, did not tweak the non Windows functions but added TODO's to move them to the new functions.
Readded the logic for Windows addresses that was removed with the addition of the non Windows logic. The existing logic caused the registry entry to flap and had the agent try to add the address every other run as a result. 
Updated the Windows package install to use the new versioning scheme, also updated the install script to match what we have done with osconfig


google_guest_agent/addresses.go
```diff
@@ -112,7 +114,41 @@ func getInterfaceByMAC(mac string, ifs []net.Interface) (net.Interface, error) {
...
+
+func getIPForwardEntries() ([]ipForwardEntry, error) {
+	if runtime.GOOS == "windows" {
+		return getIPForwardTable()
+	}
+
+	return nil, errors.New("getRoutes unimplemented on non Windows systems")
+}
+
+func addIPForwardEntry(route ipForwardEntry) error {
```
hopkiw: this function is createipforwardentry + os check. is there a benefit to maintaining that separation?

adjackura: In the future I want to remove addLocalRoute (which is linux only) and use this function for both systems. I think having the switch here makes the manager main code easier to parse, an alternative would be to have the function gated by windows and non windows build tags (that would remove the need for the non windows stub entry)

hopkiw: I think we should stick to only putting code in build-tag restricted files if it only builds for that platform. Stubs are the lesser evil here.

If this is the future primary function, let's just add TODO like your others to signify that intent to readers, and can go forward with this code as is.

hopkiw: discussed irl and reversed that, actually. functions that are fundamentally OS-specific can just have different OS implementations

PR #7, opened 2019-10-23
Fix Windows account manager and other bugs

adjackura:
Windows reset password flow now ensures the user is in the Admin group
Actually unmarshal windows-keys, updated tests
Updated builds to work as expected

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/guest-agent
 
-go 1.13
+go 1.12
 
 require (
 	github.com/GoogleCloudPlatform/guest-logging-go v0.0.0-20191018162143-d6c1e38c4e8f
--- a/google_guest_agent/addresses_unix.go
+++ b/google_guest_agent/addresses_unix.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"errors"
+	"net"
 )
 
 // TODO: addLocalRoute and addRoute should be merged with the addition of ipForwardType to ipForwardEntry.
@@ -28,4 +29,12 @@ func addIPForwardEntry(route ipForwardEntry) error {
 // TODO: getLocalRoutes and getIPForwardEntries should be merged.
 func getIPForwardEntries() ([]ipForwardEntry, error) {
 	return nil, errors.New("getIPForwardEntries unimplemented on non Windows systems")
+}
+
+func addAddress(ip net.IP, mask net.IPMask, index uint32) error {
+	return errors.New("addAddress unimplemented on non Windows systems")
+}
+
+func removeAddress(ip net.IP, index uint32) error {
+	return errors.New("removeAddress unimplemented on non Windows systems")
 }
--- a/google_guest_agent/metadata.go
+++ b/google_guest_agent/metadata.go
@@ -15,8 +15,10 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strconv"
@@ -39,6 +41,39 @@ var (
 type metadata struct {
 	Instance instance
 	Project  project
+}
+
+func (m *metadata) UnmarshalJSON(b []byte) error {
+	// We can't unmarshal into metadata directly as it would create an infinite loop.
+	type temp metadata
+	var t temp
+	err := json.Unmarshal(b, &t)
+	if err == nil {
+		*m = metadata(t)
+		return nil
+	}
+
+	// If this is a syntax error return a useful error.
+	sErr, ok := err.(*json.SyntaxError)
+	if !ok {
+		return err
+	}
+
+	// Byte number where the error line starts.
+	start := bytes.LastIndex(b[:sErr.Offset], []byte("\n")) + 1
+	// Assume end byte of error line is EOF unless this isn't the last line.
+	end := len(b)
+	if i := bytes.Index(b[start:], []byte("\n")); i >= 0 {
+		end = start + i
+	}
+
+	// Position of error in line (where to place the '^').
+	pos := int(sErr.Offset) - start
+	if pos != 0 {
+		pos = pos - 1
+	}
+
+	return fmt.Errorf("JSON syntax error: %s \n%s\n%s^", err, b[start:end], strings.Repeat(" ", pos))
 }
 
 type virtualClock struct {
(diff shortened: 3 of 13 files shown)


google_guest_agent/addresses_unix.go
```diff
@@ -29,3 +30,11 @@ func addIPForwardEntry(route ipForwardEntry) error {
 func getIPForwardEntries() ([]ipForwardEntry, error) {
 	return nil, errors.New("getIPForwardEntries unimplemented on non Windows systems")
 }
+
+func addAddress(ip net.IP, mask net.IPMask, index uint32) error {
+	return nil
```
hopkiw: is this a stub? why do other stubs return errors but this returns nil?

google_guest_agent/metadata.go
```diff
@@ -230,5 +233,32 @@ func getMetadata(ctx context.Context, hang bool) (*metadata, error) {
...
 	}
 	var ret metadata
-	return &ret, json.Unmarshal(md, &ret)
+	if err := json.Unmarshal(md, &ret); err != nil {
+		return nil, jsonError(md, err)
+	}
+	return &ret, nil
+}
+
+func jsonError(data []byte, err error) error {
```
hopkiw: it seems like this should be incorporated into our unmarshal rather than invoked optionally when handling errors. but i like the function!

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/guest-agent
 
-go 1.13
+go 1.12
```
hopkiw: can you comment on the downgrade?

google_guest_agent/non_windows_accounts.go
```diff
@@ -340,25 +338,16 @@ func createGoogleUser(user string) error {
 // permissions are removed but the user remains on the system. Group membership
 // is not changed.
 func removeGoogleUser(user string) error {
-	var err error
 	if config.Section("Accounts").Key("deprovision_remove").MustBool(true) {
 		userdel := config.Section("Accounts").Key("userdel_cmd").MustString("userdel -r {user}")
```
hopkiw: one thing to note is that by moving functions into per-platform variants as you have, that you're not doing so for everything. so here we had a consistency between adding and removing users that was easy to see - both built their local shell command and then executed it. now we have add user which calls a platform-dependent function, and then removeuser which is doing the local building of command. i am not sure what we gain by having createUser be a consistent name for a function

packaging/build_packages_wf.json
```diff
@@ -108,14 +112,15 @@
             {
               "initializeParams": {
                 "diskType": "pd-ssd",
-                "sourceImage": "projects/rhel-cloud/global/images/family/rhel-8"
+                "sourceImage": "projects/centos-cloud/global/images/family/centos-8"
```
hopkiw: why this switch? they're somewhat dissimilar atm

packaging/debian/changelog
```diff
@@ -1,4 +1,4 @@
-google-guest-agent (1:20191009.00-g1) stable; urgency=medium
+google-guest-agent (1:20191021.00-g1) stable; urgency=medium
```
hopkiw: will allow this...this time >:)

google_guest_agent/addresses_unix.go
adjackura: Copy paste of an existing function and I didn't add the error like i did for others. I can add it now if you want

google_guest_agent/metadata.go
adjackura: This was the quickest way to get this function in to log errors, its a modification of what I do with daisy

go.mod
adjackura: This was to match what we had elsewhere, wanted to keep the same for osconfig agent. If we want to move to 1.13 I'd be happy to but want to make sure it works for our partners

google_guest_agent/non_windows_accounts.go
adjackura: This one was tough, I mainly moved it because these had windows comparisons and the remove one didn't (we can add it like we've done elsewhere though). One thing I want to do is ensure there is no misunderstanding in the code, the fact that unix only functions are in the same namespace as the windows build without a clear naming scheme to differentiate is worrisome.

google_guest_agent/addresses_unix.go
hopkiw: yeah let's be consistent there, at least for readability

google_guest_agent/metadata.go
hopkiw: can you add TODO if it's too much work to put this in return of Unmarshal() ?

packaging/build_packages_wf.json
adjackura: Speed. I found building on centos was much faster, at least in the past (also cheaper)

go.mod
hopkiw: ack

packaging/debian/changelog
adjackura: Yeah I know... :p this will mostly be irrelevant when we have auto versioning I hope

google_guest_agent/non_windows_accounts.go
hopkiw: OK, let's discuss further but can leave for now. I think where there is a common behavior but with platform-dependent implementation, esp if one or both can only be compiled on a platform, we can do this. Otherwise just mentioning in naming is probably better and easier to manage in the long run.

google_guest_agent/metadata.go
```diff
@@ -41,6 +43,39 @@ type metadata struct {
...
+	// We can't unmarshal into metadata directly as it would create an infinite loop.
+	type temp metadata
+	var t temp
+	err := json.Unmarshal(b, &t)
+	if err == nil {
+		*m = metadata(t)
+		return nil
+	}
+
+	// If this is a syntax error return a useful error.
```
hopkiw: nit: the comment describes the opposite of what we're doing here

google_guest_agent/metadata.go
```diff
@@ -230,5 +266,8 @@ func getMetadata(ctx context.Context, hang bool) (*metadata, error) {
 		return nil, err
 	}
 	var ret metadata
-	return &ret, json.Unmarshal(md, &ret)
+	if err := json.Unmarshal(md, &ret); err != nil {
+		return nil, err
```
hopkiw: Why not:
```
err := json.Unmarshal(md, &ret)
return &ret, err
```

adjackura: reverted as we dont do the jsonerr herr anymore

PR #9, opened 2019-11-13
Linux instance setup

hopkiw:


google_guest_agent/instance_setup.go
```diff
@@ -75,6 +91,181 @@ func agentInit() error {
...
+	defer file.Close()
+
+	file.Write([]byte(header))
+
+	_, err = boto.WriteTo(file)
+	return err
+}
+
+func writeInstanceID(id string) error {
+	file, err := os.Create("/etc/instance_id")
```
adjackura: ioutil.Writefile ?

adjackura: If not you'll need to call Close and check the error

google_guest_agent/instance_setup.go
```diff
@@ -75,6 +90,142 @@ func agentInit() error {
 			}
 			break
 		}
+	} else {
+		// Check if instance ID has changed, and if so, consider this
+		// the first boot of the instance.
+		instance_id, err := ioutil.ReadFile("/etc/instance_id")
```
adjackura: Can you add a TODO in the Windows comment above to make this section cross platform. There's a number of one time instance setup things I'd like to remove from instance_setup.ps1

PR #10, opened 2019-11-18
Instance and boto configs

hopkiw:
For instance_configs.cfg:
Ship a flat text config and, if not already present in /etc/default, copy there.
Do this instead of shipping directly in the package, as that will mark it as a conffile and prompt users to confirm all changes. We intentionally give up the ability to update existing configs in order not to confuse or bother users with this new behavior. Users can now edit the main config directly, as omissions are safe (we define a default in code when we use a config).

For boto.cfg:
Would be great to do the same as instance_configs.cfg, but this file must have the numeric project ID, so it has to be generated. Do this in the one-time section i.e. "instance setup".

### DIFF
--- a/google_guest_agent/metadata.go
+++ b/google_guest_agent/metadata.go
@@ -97,8 +97,9 @@ type networkInterfaces struct {
 }
 
 type project struct {
-	Attributes attributes
-	ProjectID  string
+	Attributes       attributes
+	ProjectID        string
+	NumericProjectID string
 }
 
 type attributes struct {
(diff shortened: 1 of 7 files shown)


hopkiw:
/hold

this is writing case-insensitive, will fix that and unhold.

google_guest_agent/metadata.go
```diff
@@ -242,11 +243,12 @@ func getMetadata(ctx context.Context, hang bool) (*metadata, error) {
 	}
 	finalURL += ("&last_etag=" + etag)
 
-	req, err := http.NewRequestWithContext(ctx, "GET", finalURL, nil)
+	req, err := http.NewRequest("GET", finalURL, nil)
```
adjackura: Why the switch back from WithContext?

hopkiw: not supported in the go version we use in builds, apparently.

PR #16, opened 2019-12-12
Fixes from debugging sessions

hopkiw:
More nuanced runCmd functions
Create boto.cfg, handle legacy instance_id in config
EnableIPv6 function
Correct SSH host key generation logic

### DIFF
--- a/google_guest_agent/instance_setup.go
+++ b/google_guest_agent/instance_setup.go
@@ -178,7 +178,7 @@ func generateSSHKeys() error {
 	}
 
 	// Generate new keys and upload to guest attributes.
-	for keytype, _ := range keytypes {
+	for keytype := range keytypes {
 		keyfile := fmt.Sprintf("/etc/ssh/ssh_host_%s_key", keytype)
 		if err := runCmd(exec.Command("ssh-keygen", "-t", keytype, "-f", keyfile+".temp", "-N", "", "-q")); err != nil {
 			logger.Warningf("Failed to generate SSH host key %q: %v", keyfile, err)
--- a/google_guest_agent/instance_setup.go
+++ b/google_guest_agent/instance_setup.go
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	instance_id_file = "/etc/google_instance_id"
+	instanceIDFile = "/etc/google_instance_id"
 )
 
 func forwardEntryExists(fes []ipForwardEntry, fe ipForwardEntry) bool {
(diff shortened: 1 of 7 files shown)


google_guest_agent/main.go
```diff
@@ -102,7 +102,7 @@ func runUpdate() {
 	mgrs := []manager{&addressMgr{}}
 	switch runtime.GOOS {
 	case "windows":
-		mgrs = append(mgrs, []manager{newWsfcManager(), &winAccountsMgr{}, &diagnosticsMgr{}}...)
+		mgrs = append(mgrs, []manager{newWsfcManager(), &winAccountsMgr{}}...)
```
adjackura: Add diagnosticsMgr back in

google_guest_agent/main.go
```diff
@@ -183,35 +183,52 @@ func run(ctx context.Context) {
...
+func runCmdOutput(cmd *exec.Cmd) *execResult {
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
 	err := cmd.Run()
 	if err != nil {
 		if ee, ok := err.(*exec.ExitError); ok {
-			return fmt.Errorf(string(ee.Stderr))
+			return &execResult{code: ee.ExitCode(), err: stderr.String()}
```
adjackura: For future proofing I would add stdout here as well, just because its non-zero doesn't mean there stderr

adjackura: Actually, just always return &execResult{code: ee.ExitCode(), out: stdout.String(), err: stderr.String()}, it'll make it cleaner.

google_guest_agent/instance_setup.go
```diff
@@ -158,35 +156,61 @@ func generateSSHKeys() error {
...
 	// Generate new keys and upload to guest attributes.
-	keyTypes := config.Section("InstanceSetup").Key("host_key_types").MustString("ecdsa,ed25519,rsa")
-	for _, keyType := range strings.Split(keyTypes, ",") {
-		outfile := fmt.Sprintf("/etc/ssh/ssh_host_%s_key", keyType)
-		if err := runCmd(exec.Command("ssh-keygen", "-t", keyType, "-f", outfile, "-N", "", "-q")); err != nil {
-			return fmt.Errorf("Failed to generate SSH host key %q", outfile)
+	for keytype := range keytypes {
+		keyfile := fmt.Sprintf("/etc/ssh/ssh_host_%s_key", keytype)
+		if err := runCmd(exec.Command("ssh-keygen", "-t", keytype, "-f", keyfile+".temp", "-N", "", "-q")); err != nil {
+			logger.Warningf("Failed to generate SSH host key %q: %v", keyfile, err)
```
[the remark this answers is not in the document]
hopkiw: this is intentional. we move it into final location below on line 195

google_guest_agent/main.go
hopkiw: i don't know what you mean by 'always'. this can only be done if the error was type `*exec.ExitError`, otherwise there is no ee. so you mean only to change this one line, right?

google_guest_agent/instance_setup.go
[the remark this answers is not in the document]
adjackura: I mean for the log line, the file name does not match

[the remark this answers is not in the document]
hopkiw: we log about our intent to create the final file, this was also intentional.

PR #18, opened 2020-01-06
Script runner

hopkiw:
* Copy metadata runner from GoogleCloudPlatform/compute-image-windows
* Update RPM scriptlets and debian maintainer scripts
* Copy windows packaging and files to new locations
* Add config file support (step through commits to see these changes)

The windows packages will remain split as they are now, but the script runner will be shipped as part of the `google-guest-agent` package on linux.

### DIFF
--- a/packaging/google-guest-agent.spec
+++ b/packaging/google-guest-agent.spec
@@ -29,6 +29,8 @@ BuildArch: %{_arch}
 BuildRequires: systemd
 %endif
 
+Obsoletes: python-google-compute-engine, python3-google-compute-engine
+
 %description
 Contains the Google guest agent binary.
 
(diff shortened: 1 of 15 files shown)


hopkiw:
/hold
i want both andrew and zach to approve here - andrew has script runner context, zach has RPM/debian packaging context

packaging/google-guest-agent.spec
```diff
@@ -29,58 +29,134 @@ BuildArch: %{_arch}
 BuildRequires: systemd
 %endif
 
+Obsoletes: python-google-compute-engine, python3-google-compute-engine
```
zmarano: Hmm... We may want to consider doing something else here. We may still have a python package (as a library) around.

hopkiw: understand this concern, but if we don't remove that package the legacy daemons won't be stopped and deregistered. i think we should update python-gce to stop producing the entrypoint scripts and upload that to pypi for anyone who needs it.

zmarano: Yes. I think that makes sense.

hopkiw:
ah, missing extra_refs for the packagebuild setup. will run that again once oss-test-infra#213 is merged

hopkiw:
the config was definitely updated, but the packagebuilder isn't being checked out.. try once more, but maybe there's a caching issue. would like packagebuilder to succeed before merging since this is a change to packaging

/packagebuild

hopkiw:
updated build to perform googet builds as well

/packagebuild

PR #19, opened 2020-01-08
wait on metadata indefinitely

hopkiw:
wait indefinitely on metadata **for linux**


google_guest_agent/instance_setup.go
```diff
@@ -96,6 +98,15 @@ func agentInit() error {
 			break
 		}
 	} else {
+		// The below actions require metadata to be set, so if it
+		// hasn't yet been set, wait on it here. In instances without
+		// network access, this will become an indefinite wait.
+		// TODO: split agentInit into needs-network and no-network functions.
+		for newMetadata == nil {
+			newMetadata, _ = getMetadata(ctx, false)
+			time.Sleep(1 * time.Second)
```
adjackura: This seems aggressive, existing process in script-runner has back off to increase sleep time up to a point.

hopkiw: backoff retries make sense when you don't want to overload a remote service or spam something to logs, but this will be an indefinite wait with nothing else being performed, why add complexity?

PR #20, opened 2020-01-10
Instance config

hopkiw:
Add support for remaining instance config options
Deprecate the keys `usermod_cmd` and `ip_forwarding_daemon` in instance config


hopkiw:
@illfelder please tal

PR #22, opened 2020-01-17
Add guest-aware snapshot listener

maxnelso:
This listener connects to a gRPC server running on the host and waits for "pre snapshot requests" and "post snapshot requests". On these requests, the listener will run scripts as defined in instance_configs.cfg and send back the result to the host.

Note that this adds a a few snapshot related configuration fields to instance_configs.cfg, so the documentation at: https://github.com/GoogleCloudPlatform/compute-image-packages/tree/master/packages/python-google-compute-engine should be updated.


google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+type invalidSnapshotConfig struct {
+	msg string
+}
+
+func (e *invalidSnapshotConfig) Error() string {
+	return fmt.Sprintf("invalid config: %s", e.msg)
+}
+
+func getSnapshotConfig() (snapshotConfig, error) {
+	setConfig()
```
hopkiw: prevents this function from being used with any other config context. we set the config as a global to allow functions to use it however it has been set externally

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+	var conf snapshotConfig
+	conf.Timeout = config.Section("Snapshots").Key("timeout_in_seconds").MustInt(60)
+	conf.ContinueOnScriptError = config.Section("Snapshots").Key("continue_on_script_error").MustBool(false)
+	conf.PreSnapshotScriptURL = config.Section("Snapshots").Key("pre_snapshot_script").MustString("")
```
hopkiw: don't use `.MustString("")`, just use `.String()`

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+	"syscall"
+	"time"
+
+	pb "github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/protos"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	"github.com/hashicorp/golang-lru"
+	"google.golang.org/grpc"
+)
+
+const testPreSnapshotScript = "preSnapshotScript.sh"
```
hopkiw: define test values in test files

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+	conf.Timeout = config.Section("Snapshots").Key("timeout_in_seconds").MustInt(60)
+	conf.ContinueOnScriptError = config.Section("Snapshots").Key("continue_on_script_error").MustBool(false)
+	conf.PreSnapshotScriptURL = config.Section("Snapshots").Key("pre_snapshot_script").MustString("")
+	conf.PostSnapshotScriptURL = config.Section("Snapshots").Key("post_snapshot_script").MustString("")
+
+	if conf.PreSnapshotScriptURL == "" && conf.PostSnapshotScriptURL == "" {
+		msg := "neither pre or post snapshot script has been configured"
+		logger.Errorf("%s", msg)
+		return conf, &invalidSnapshotConfig{msg}
+	}
```
hopkiw: just to be explicit with above feedback, this should be either logged or returned with the message defined inline

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return conf, nil
+}
+
+func runScript(path string, disks string, timeout int) (int, error) {
```
hopkiw: np `path, disks string, timeout int`

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+	logger.Infof("running script at: %s", path)
```
hopkiw: need more context on this if we're going to log it, something like 'running pre-snapshot script %q'

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+	case err := <-done:
+		if err != nil {
+			exiterr := err.(*exec.ExitError)
+			status := exiterr.Sys().(syscall.WaitStatus)
+			logger.Errorf("error while running script at %s, return with exit code %d", path, status.ExitStatus())
+			return status.ExitStatus(), nil
+		}
+	}
+	return 0, nil
+}
```
hopkiw: this section should be simply a call to one of our existing runcmd functions. if there's a missing feature you need there (seems like timeout, maybe) then let's consider how we can extend.

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+
+	if !config.ContinueOnScriptError && ec != 0 {
+		logger.Infof("script at: %s failed with error code %d", path, ec)
+		return pb.AgentErrorCode_UNHANDLED_SCRIPT_ERROR, ec
+	}
+
+	return pb.AgentErrorCode_NO_ERROR, ec
+}
+
+func listenForSnapshotRequests(address string, requestChan chan *pb.GuestMessage) {
```
hopkiw: is this meant to be a bidirectional channel?

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+		r, err := c.CreateConnection(ctx, &guestReady)
+		if err != nil {
+			logger.Errorf("error creating connection: %v", err)
+			continue
+		}
+		logger.Infof("created hanging connection with snapshot service")
+		for {
+			request, err := r.Recv()
+			if err != nil {
+				logger.Errorf("error reading request: %v", err)
```
hopkiw: error reading snapshot request

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+		ctx, cancel := context.WithCancel(context.Background())
+		guestReady := pb.GuestReady{
+			RequestServerInfo: false,
+		}
+		r, err := c.CreateConnection(ctx, &guestReady)
+		if err != nil {
+			logger.Errorf("error creating connection: %v", err)
+			continue
+		}
+		logger.Infof("created hanging connection with snapshot service")
```
hopkiw: unsure this needs to be logged

google_guest_agent/protos/snapshot_service.proto
```diff
@@ -0,0 +1,99 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
```
hopkiw: do we need to check the proto file in here..?

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+			logger.Infof("received snapshot request")
+			requestChan <- request
+		}
+		cancel()
+	}
+}
+
+func getSnapshotResponse(guestMessage *pb.GuestMessage) *pb.SnapshotResponse {
+	switch request := guestMessage.Msg.(type) {
```
hopkiw: consider a paradigm we use for this kind of test in osconfig: https://github.com/GoogleCloudPlatform/osconfig/blob/master/policies/recipes/installrecipe.go#L89

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+			return response
+
+		}
+
+		var url string
+		switch request.SnapshotRequest.GetType() {
+		case pb.OperationType_PRE_SNAPSHOT:
+			logger.Infof("handling pre snapshot request for operation id %d", request.SnapshotRequest.GetOperationId())
+			if seenPreSnapshotoperationIds.Contains(request.SnapshotRequest.GetOperationId()) {
+				logger.Infof("duplicate pre snapshot request operation id %d", request.SnapshotRequest.GetOperationId())
```
hopkiw: does this need to be logged..?

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+			logger.Errorf("unhandled operation type %d", request.SnapshotRequest.GetType())
+			return nil
+		}
+
+		agentErrorCode, scriptsReturnCode := handleRequest(config, url, request.SnapshotRequest.GetDiskList())
+		response.ScriptsReturnCode = int32(scriptsReturnCode)
+		response.AgentReturnCode = agentErrorCode
+
+		return response
+	case *pb.GuestMessage_ServerInfo:
```
hopkiw: why blank case?

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+	}
+
+}
+
+func startSnapshotListener(testMode bool, testSnapshotPort int) {
+	requestChan := make(chan *pb.GuestMessage)
+	var address string
+	if testMode {
+		address = testIP + ":" + strconv.Itoa(testSnapshotPort)
```
hopkiw: `fmt.Sprintf` seems friendlier

hopkiw: this is also probably obviated by the feedback on test constants

google_guest_agent/snapshot_listener.go
maxnelso: Think this might be helpful for our debugging if that's alright.

google_guest_agent/snapshot_listener.go
maxnelso: Silliness, made default.

google_guest_agent/snapshot_listener.go
maxnelso: See if this is what you had in mind.

google_guest_agent/snapshot_listener.go
maxnelso: Silliness, removed.

google_guest_agent/snapshot_listener.go
maxnelso: Holding off on this for now

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,252 @@
...
+	PreSnapshotScriptUrl  string
+	PostSnapshotScriptUrl string
+	Enabled               bool
+}
+
+type scriptsTimeoutError struct {
+	path string
+}
+
+func (e *scriptsTimeoutError) Error() string {
```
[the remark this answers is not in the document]
hopkiw: i'm saying this is something you shouldn't do - you're passing information around embedded in the type instead of doing so explicitly in some way.

google_guest_agent/snapshot_listener.go
hopkiw: then shouldn't it be logger.Debugf with Debug set to true in your testing builds? what i meant by this question is does it need to be logged on customer instances

google_guest_agent/main.go
```diff
@@ -272,12 +282,11 @@ func main() {
...
 	}
 
-	var action string
-	if len(os.Args) < 2 {
-		action = "run"
-	} else {
-		action = os.Args[1]
-	}
+	flag.StringVar(&action, "action", "run", "the action to run")
+	flag.IntVar(&snapshotServiceTestPort, "snapshot_service_test_port", 0, "the test snapshot service port")
```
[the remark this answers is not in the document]
hopkiw: no, both the spec and the build receive the version variable, it is updated during the build as a linker argument. you can see it here
https://github.com/GoogleCloudPlatform/guest-agent/blob/master/packaging/google-guest-agent.spec#L43

google_guest_agent/protos/snapshot_service.proto
hopkiw: i don't have any context for doing this, it seems harmless but also useless. all of our other proto work relies on published protos

google_guest_agent/snapshot_listener.go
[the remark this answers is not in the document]
maxnelso: Hm ok, tried to refactor this so it doesn't follow this pattern, see if you like this better.

google_guest_agent/main.go
```diff
@@ -127,6 +131,7 @@ func run(ctx context.Context) {
 		opts.FormatFunction = logFormat
 		opts.Writers = []io.Writer{&serialPort{"COM1"}}
 	}
+	opts.Writers = append(opts.Writers, os.Stdout)
```
hopkiw: we still don't want this

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,239 @@
...
+func getSnapshotConfig() (snapshotConfig, error) {
+	cfgfile := configPath
+	if runtime.GOOS == "windows" {
+		cfgfile = winConfigPath
+	}
+
+	config, err := parseConfig(cfgfile)
+	if err != nil && !os.IsNotExist(err) {
+		logger.Errorf("Error parsing config %s: %s", cfgfile, err)
+	}
```
hopkiw: these are general configs and not snapshot-specific, should be handled in main as before and these functions should assume this part is done

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,239 @@
...
+
+		if config.ContinueOnScriptError {
+			agentErrorCode = pb.AgentErrorCode_NO_ERROR
+		}
+
+		return scriptsReturnCode, agentErrorCode
+	}
+
+	done := make(chan error)
+	go func() { done <- cmd.Wait() }()
```
hopkiw: what is this doing?

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,239 @@
...
+
+	select {
+	case <-timer:
+		logger.Errorf("timeout while running script at %s", path)
+		// Exit code of the timeout command when it times out
+		scriptsReturnCode, agentErrorCode = 124, pb.AgentErrorCode_SCRIPT_TIMED_OUT
+	case err := <-done:
+		if err != nil {
+			exiterr := err.(*exec.ExitError)
+			status := exiterr.Sys().(syscall.WaitStatus)
```
hopkiw: we have exec wrappers, you shouldn't have to do this here

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if config.ContinueOnScriptError {
+		agentErrorCode = pb.AgentErrorCode_NO_ERROR
+	}
+
+	return scriptsReturnCode, agentErrorCode
+}
+
+func listenForSnapshotRequests(address string, requestChan chan<- *pb.GuestMessage) {
+	for {
```
hopkiw: you have an inner loop for r.Recv(), why nest the loops?

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,239 @@
...
+		response.AgentReturnCode = agentErrorCode
+
+		return response
+	default:
+	}
+	return nil
+}
+
+func handleSnapshotRequests(address string, requestChan <-chan *pb.GuestMessage) {
+	for {
```
hopkiw: same qn here - why are you infinitely creating new connections? can't handler dial up connections only when there are messages on the channel?

google_guest_agent/main.go
maxnelso: Removed!

google_guest_agent/snapshot_listener.go
maxnelso: The idea was on broken connections we should attempt to reconnect.

google_guest_agent/snapshot_listener.go
maxnelso: Same thing, was worried about reconnecting.

google_guest_agent/main.go
```diff
@@ -247,6 +251,38 @@ func runCmdOutput(cmd *exec.Cmd) *execResult {
 	return &execResult{code: 0, out: stdout.String()}
 }
 
+func runCmdOutputAndTimeout(cmd *exec.Cmd, timeoutSec int) *execResult {
```
hopkiw: i think this can instead be a function that wraps `runCmdOutput`, but passes in an `exec.CommandContext` instead. it will avoid the need to replicate so much code. see: http://golang.org/pkg/os/exec/#CommandContext

hopkiw: there haven't been any pushes since this was resolved, so unresolving

maxnelso: Weird, I might be losing my mind, try now!

google_guest_agent/main.go
```diff
@@ -247,6 +252,16 @@ func runCmdOutput(cmd *exec.Cmd) *execResult {
 	return &execResult{code: 0, out: stdout.String()}
 }
 
+func runCmdOutputAndTimeout(timeoutSec int, name string, args ...string) *execResult {
```
adjackura: runCmdOutputWithTimeout

google_guest_agent/main.go
```diff
@@ -270,12 +285,11 @@ func closer(c io.Closer) {
 
 func main() {
 	ctx := context.Background()
-	var action string
-	if len(os.Args) < 2 {
-		action = "run"
-	} else {
-		action = os.Args[1]
-	}
+	flag.StringVar(&action, "action", "run", "the action to run")
```
adjackura: Please don't add flags to this binary, it's not meant to be run on the cmdline and we have existing ways users are expected to configure it, either via metadata or config file

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,196 @@
...
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	pb "github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/protos"
```
adjackura: make the library github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/snapshot_service
and the import will be:
sspb "github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/snapshot_service"

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,196 @@
...
+	"google.golang.org/grpc"
+)
+
+var (
+	seenPreSnapshotOperationIds, _  = lru.New(128)
+	seenPostSnapshotOperationIds, _ = lru.New(128)
+)
+
+type snapshotConfig struct {
+	Timeout               int // seconds
```
adjackura: nit: these should be private fields as the struct is private (unless there is some need like the json marshaller)

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,196 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	pb "github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/protos"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	"github.com/hashicorp/golang-lru"
```
adjackura: Avoid third party imports as much as possible, you should be able to implement this with the sync or sync/atomic package correct?

google_guest_agent/main.go
maxnelso: Fixed!

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,202 @@
...
+	"google.golang.org/grpc"
+)
+
+var (
+	seenPreSnapshotOperationIds  = lru.New(128)
+	seenPostSnapshotOperationIds = lru.New(128)
+)
+
+type snapshotConfig struct {
+	timeout               int // seconds
```
adjackura: time.Duration?

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,202 @@
...
+	msg string
+}
+
+func (e *invalidSnapshotConfig) Error() string {
+	return fmt.Sprintf("invalid config: %s", e.msg)
+}
+
+func getSnapshotConfig() (snapshotConfig, error) {
+	var conf snapshotConfig
+	conf.timeout = config.Section("Snapshots").Key("timeout_in_seconds").MustInt(60)
```
adjackura: would you prefer strings and use time.ParseDuration()? up to you, int with name of timeout_in_seconds is also not bad assuming this is always a few seconds

google_guest_agent/snapshot_listener.go
```diff
@@ -0,0 +1,202 @@
...
+			logger.Infof("handling post snapshot request for operation id %d", request.GetOperationId())
+			_, found := seenPostSnapshotOperationIds.Get(request.GetOperationId())
+			if found {
+				logger.Infof("duplicate post snapshot request operation id %d", request.GetOperationId())
+				return nil
+			}
+			seenPostSnapshotOperationIds.Add(request.GetOperationId(), request.GetOperationId())
+			url = config.postSnapshotScriptURL
+		default:
+			logger.Errorf("unhandled operation type %d", request.GetType())
```
adjackura: Log lines should be capitalized and end with punctuation (here and elsewhere)

go.sum
```diff
@@ -106,6 +103,7 @@ github.com/kisielk/gotool v1.0.0/go.mod h1:XhKaO+MFFWcvkIS/tQcRk01m1F5IRFswLeQ+o
 github.com/kr/pretty v0.1.0/go.mod h1:dAy3ld7l9f0ibDNOQOHHMYYIIbhfbHSm3C4ZsoJORNo=
 github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
+github.com/maxnelso/guest-agent v0.0.0-20200109220557-df783fbc67b6 h1:2ZPXooTppQRlCkJO7nbVdXPugJMGZZhIyLRtW3EXTSk=
```
adjackura: make sure you run 'go mod tidy'

google_guest_agent/snapshot_listener.go
maxnelso: Fixed!

google_guest_agent/snapshot_listener.go
maxnelso: Think our documentation for this is already written, probably best not to change at this point.

go.sum
maxnelso: Ahh thank you I didn't know about this.

google_guest_agent/snapshot_listener.go
maxnelso: Fixed!

PR #25, opened 2020-01-27
Fix upstart,selinux

hopkiw:
* Change upstart init to not rely on google-instance-setup (deprecated)
* Instance ID comparisons should strip newlines
* Run restorecon on authorized_keys files

### DIFF
--- a/google_guest_agent/instance_setup.go
+++ b/google_guest_agent/instance_setup.go
@@ -157,7 +157,7 @@ func agentInit(ctx context.Context) error {
 					logger.Warningf("Failed to write instance ID file: %v", err)
 				}
 			}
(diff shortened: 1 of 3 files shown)


google_guest_agent/instance_setup.go
```diff
@@ -157,7 +157,7 @@ func agentInit(ctx context.Context) error {
 					logger.Warningf("Failed to write instance ID file: %v", err)
 				}
 			}
-			if newMetadata.Instance.ID.String() != string(instanceID) {
+			if newMetadata.Instance.ID.String() != strings.TrimSuffix(string(instanceID), "\n") {
```
adjackura: Think about using https://golang.org/pkg/strings/#TrimSpace here, assuming you trust what you get from metadata

PR #26, opened 2020-01-28
amend upstart again to be before hostkeys

hopkiw:
this is actually equivalent to what legacy instance setup depended on, and has an impact on hostkey generation. to avoid any surprise to users, match what we had before.


hopkiw:
on second thought, i think we should be OK with this slight change to startup ordering, and the new way is better since it doesn't tie us to rsyslog restarts

PR #29, opened 2020-02-06
Avoid setting route on Teredo Pseudo-Interface

rofuentes:
2020/02/06 04:27:25 GCEGuestAgent: GCE Agent Started (version 20200129.00)
2020/02/06 04:27:25 GCEGuestAgent: Adding route to metadata server on "Teredo Tunneling Pseudo-Interface" (index: 3)
2020/02/06 04:27:25 GCEGuestAgent: Error adding route to metadata server on "Teredo Tunneling Pseudo-Interface" (index: 3): nonzero return code from CreateIpForwardEntry: Element not found.

### DIFF
--- a/google_guest_agent/instance_setup.go
+++ b/google_guest_agent/instance_setup.go
@@ -75,7 +75,8 @@ func agentInit(ctx context.Context) error {
 
 		for _, iface := range interfaces {
 			// Only take action on interfaces that are up, are not Loopback, and are not vEtherent (commonly setup by docker).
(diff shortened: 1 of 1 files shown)


adjackura:
I think we may want to handle this based on adapter description, a better way may be to set this based on routing but thats more complicated and requires more calls. For no using windows.GetAdaptersAddresses will give you the description.

rofuentes:
This may be fixed by https://github.com/GoogleCloudPlatform/guest-agent/pull/47

PR #30, opened 2020-02-19
always restart services

hopkiw:
there are a growing number of edge cases where services aren't restarted but need to be. we had been conservative about service restarts, but really the current set are all very resilient to restarts so let's always restart when a change to oslogin settings is detected.


hopkiw:
comment to kick presubmits, hopefully

PR #36, opened 2020-04-08
correct SSH key handling

hopkiw:
[Public GCE docs](https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys#sshkeyformat) reference the following as a valid SSH key format:
```
ssh-rsa [KEY_VALUE] [USERNAME]
```

However, the legacy guest environment supports also:
```
ssh-rsa [KEY_VALUE]
```

Correct the filtering done in `removeExpiredKeys()` not to skip keys in this format. Update tests to accomodate it.

### DIFF
--- a/google_guest_agent/accounts_test.go
+++ b/google_guest_agent/accounts_test.go
@@ -202,10 +202,12 @@ func TestRemoveExpiredKeys(t *testing.T) {
 	}{
 		{`user:ssh-rsa [KEY] google-ssh {"userName":"user@email.com", "expireOn":"2028-11-08T19:30:47+0000"}`, true},
 		{`user:ssh-rsa [KEY] google-ssh {"userName":"user@email.com", "expireOn":"2028-11-08T19:30:47+0700"}`, true},
+		{`user:ssh-rsa [KEY] google-ssh {"userName":"user@email.com", "expireOn":"2028-11-08T19:30:47+0700", "futureField": "UNUSED_FIELDS_IGNORED"}`, true},
 		{`user:ssh-rsa [KEY] google-ssh {"userName":"user@email.com", "expireOn":"2018-11-08T19:30:46+0000"}`, false},
 		{`user:ssh-rsa [KEY] google-ssh {"userName":"user@email.com", "expireOn":"2018-11-08T19:30:46+0700"}`, false},
-		{`user:ssh-rsa [KEY] hostname`, true},
+		{`user:ssh-rsa [KEY] google-ssh {"userName":"user@email.com", "expireOn":"INVALID_TIMESTAMP"}`, false},
 		{`user:ssh-rsa [KEY] google-ssh`, false},
+		{`user:ssh-rsa [KEY] user`, true},
 		{`user:ssh-rsa [KEY]`, true},
 		{},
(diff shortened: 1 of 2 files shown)


google_guest_agent/accounts_test.go
```diff
@@ -206,7 +206,7 @@ func TestRemoveExpiredKeys(t *testing.T) {
 		{`user:ssh-rsa [KEY] google-ssh {"userName":"user@email.com", "expireOn":"2018-11-08T19:30:46+0700"}`, false},
 		{`user:ssh-rsa [KEY] hostname`, true},
 		{`user:ssh-rsa [KEY] google-ssh`, false},
-		{`user:ssh-rsa [KEY]`, false},
+		{`user:ssh-rsa [KEY]`, true},
```
illfelder, line 209: In the python version of the guest environment we have testing for whether to consider an SSH key expired. That testing doesn't fully match your current cases. Please make sure they align to prevent a break in backwards compatibility. Here is the existing tests for reference: https://github.com/GoogleCloudPlatform/compute-image-packages/blob/master/packages/python-google-compute-engine/google_compute_engine/accounts/tests/accounts_daemon_test.py#L118

hopkiw: thanks! all of those test cases are covered in this test or in the expiration test at top of file

illfelder: This test is a break in backwards compatibility:
```
{`user:ssh-rsa [KEY] google-ssh`, false},
```

It's also good to include some of these tests from the python guest (I don't see a parallel example in this code). Note that "False" means the key is not expired (and would be true in your test).
```
        'Invalid': False,
        'user:ssh-rsa key user@domain.com': False,
        'user:ssh-rsa key google-ssh {invalid:json}': False,
        'user:ssh-rsa key google-ssh {"userName":"user"}': False,
        'user:ssh-rsa key google-ssh {"expireOn":"invalid"}': False,
```

hopkiw: yes, that's an intentional break in backwards compatibilty, as the behavior was incorrect in the legacy guest. are there any known clients who use this behavior?

we have the test that shows users without google-ssh are considered valid, on line 207, two above the highlighted. invalid JSON is handled in the metadata test.

illfelder: A key that does not have an expire on time specified should not be considered expired. We should have a test case that covers that behavior. Ideally we should also have a test case that covers a dictionary with key/value pairs different from the fully qualified version. This is useful if ever we want to add a new parameter in the future.

Regarding the break in backwards compatibility, we don't have a way to know whether a client will break due to this change.

hopkiw: The `google-ssh` field is a reserved field that marks this as a google generated expiring key. A google generated key without an expiration blob is invalid, and won't be supported

hopkiw: For keys without `google-ssh` field, they will never be considered expired. They may be filtered in this function if they are strictly invalid i.e. contain less than 2 fields. Prior to this PR, it was 3 fields.

PR #43, opened 2020-05-26
Bugfixes

hopkiw:
3 bugfixes identified during software rollout
* IP forwarding comparison incorrectly considered '1.1.1.1' and '1.1.1.1/32' to be different
* errors calling dhclient for IPv6 would return and prevent the remainder of network setup from running
* rpm spec %preun typo caused agent not to be stopped on uninstall

behavior before IP forwarding fix:
```
$ ip route list table local type local scope host dev eth0 proto 66
local 10.138.0.104 
$ sudo systemctl restart google-guest-agent                        
$ ip route list table local type local scope host dev eth0 proto 66
$ sudo systemctl restart google-guest-agent                        
$ ip route list table local type local scope host dev eth0 proto 66
local 10.138.0.104 
```

behavior as of this PR:
```
$ ip route list table local type local scope host dev eth0 proto 66
local 10.138.0.104 
$ sudo systemctl restart google-guest-agent
$ ip route list table local type local scope host dev eth0 proto 66
local 10.138.0.104 
$ sudo systemctl restart google-guest-agent
$ ip route list table local type local scope host dev eth0 proto 66
local 10.138.0.104 
```

### DIFF
--- a/google_guest_agent/addresses.go
+++ b/google_guest_agent/addresses.go
@@ -85,16 +85,24 @@ func getForwardsFromRegistry(mac string) ([]string, error) {
 	return regFwdIPs, nil
 }
 
-func compareIPs(configuredIPs, desiredIPs []string) (toAdd, toRm []string) {
-	for _, desiredIP := range desiredIPs {
-		if !containsString(desiredIP, configuredIPs) {
-			toAdd = append(toAdd, desiredIP)
+func compareRoutes(configuredRoutes, desiredRoutes []string) (toAdd, toRm []string) {
+	for _, desiredRoute := range desiredRoutes {
+		if !strings.Contains(desiredRoute, "/") {
+			logger.Errorf("requested route %q missing mask, skipping.", desiredRoute)
+			continue
+		}
+		if !containsString(desiredRoute, configuredRoutes) {
+			toAdd = append(toAdd, desiredRoute)
 		}
 	}
 
-	for _, configuredIP := range configuredIPs {
-		if !containsString(configuredIP, desiredIPs) {
-			toRm = append(toRm, configuredIP)
+	for _, configuredRoute := range configuredRoutes {
+		if !strings.Contains(configuredRoute, "/") {
+			logger.Errorf("local route %q missing mask, skipping.", configuredRoute)
+			continue
+		}
+		if !containsString(configuredRoute, desiredRoutes) {
+			toRm = append(toRm, configuredRoute)
 		}
 	}
 	return toAdd, toRm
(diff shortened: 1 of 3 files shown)


google_guest_agent/addresses.go
```diff
@@ -143,6 +143,9 @@ func getLocalRoutes(ifname string) ([]string, error) {
 		line = strings.TrimPrefix(line, "local ")
 		line = strings.TrimSpace(line)
 		if line != "" {
+			if !strings.Contains(line, "/") {
+				line = line + "/32"
```
EricEdens, line 147: Is it possible this will introduce a regression for cases that worked prior to having the mask applied?

google_guest_agent/addresses.go
```diff
@@ -143,6 +143,9 @@ func getLocalRoutes(ifname string) ([]string, error) {
 		line = strings.TrimPrefix(line, "local ")
 		line = strings.TrimSpace(line)
 		if line != "" {
+			if !strings.Contains(line, "/") {
```
EricEdens, line 146: Is the caller of `getLocalRoutes` expecting to get IP addresses, or IP address blocks?

google_guest_agent/addresses.go
hopkiw: i can't think of a scenario that would fail, but we are also updating the tests for this, since it wasn't caught.

google_guest_agent/addresses.go
hopkiw: this function returns what the 'ip route' command outputs. if a local route has any mask other than /32, it will be present in the output.

google_guest_agent/addresses.go
```diff
@@ -85,16 +85,24 @@ func getForwardsFromRegistry(mac string) ([]string, error) {
...
-func compareIPs(configuredIPs, desiredIPs []string) (toAdd, toRm []string) {
-	for _, desiredIP := range desiredIPs {
-		if !containsString(desiredIP, configuredIPs) {
-			toAdd = append(toAdd, desiredIP)
+func compareRoutes(configuredRoutes, desiredRoutes []string) (toAdd, toRm []string) {
+	for _, desiredRoute := range desiredRoutes {
+		if !strings.Contains(desiredRoute, "/") {
+			logger.Errorf("requested route %q missing mask, skipping.", desiredRoute)
+			continue
+		}
```
EricEdens, line 93: As opposed to dropping the desiredRoute, I'd suggest just appending /32. 

We talked offline, and it sounds like your desired approach is to update all of the callers to ensure they have the mask applied to each member. This case, then, should never happen.  By adding /32 here, though, you'll at least be able to recover instead of dropping the route.

hopkiw: yeah, i don't think the compare function should modify its inputs at all. we take from metadata and from `getLocalRoutes()`, the latter of which is patched in this PR to append /32 to entries without any mask.

PR #45, opened 2020-05-27
strip /32 suffix from metadata entries

hopkiw:
This is an amendment to the fix in #43 for handling route comparisons. It turns out there are some cases where items in metadata have `/32` notation and cases where they don't (notably, forwarded IP vs alias IP). Strip these suffixes before comparing or taking actions.

### DIFF
--- a/google_guest_agent/addresses_test.go
+++ b/google_guest_agent/addresses_test.go
@@ -28,18 +28,19 @@ func TestCompareRoutes(t *testing.T) {
 	}{
 		// These should return toAdd:
 		// In Md, not present
-		{nil, []string{"1.2.3.4/32"}, []string{"1.2.3.4/32"}, nil},
-		{nil, []string{"1.2.3.4/32", "5.6.7.8/32"}, []string{"1.2.3.4/32", "5.6.7.8/32"}, nil},
+		{nil, []string{"1.2.3.4/32"}, []string{"1.2.3.4"}, nil},
+		{nil, []string{"1.2.3.4", "5.6.7.8/32"}, []string{"1.2.3.4", "5.6.7.8"}, nil},
 
 		// These should return toRm:
 		// Present, not in Md
-		{[]string{"1.2.3.4/32"}, nil, nil, []string{"1.2.3.4/32"}},
-		{[]string{"1.2.3.4/32", "5.6.7.8/32"}, []string{"5.6.7.8/32"}, nil, []string{"1.2.3.4/32"}},
+		{[]string{"1.2.3.4/32"}, nil, nil, []string{"1.2.3.4"}},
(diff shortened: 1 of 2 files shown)


google_guest_agent/addresses_test.go
```diff
@@ -28,18 +28,19 @@ func TestCompareRoutes(t *testing.T) {
 	}{
 		// These should return toAdd:
 		// In Md, not present
-		{nil, []string{"1.2.3.4/32"}, []string{"1.2.3.4/32"}, nil},
-		{nil, []string{"1.2.3.4/32", "5.6.7.8/32"}, []string{"1.2.3.4/32", "5.6.7.8/32"}, nil},
+		{nil, []string{"1.2.3.4"}, []string{"1.2.3.4"}, nil},
```
illfelder, line 31: To properly test this change, shouldn't we have test cases in the TestAddressDiff function where the metadata server returns IPs with subnetworks specified? It seems like we're missing any testing for subnetworks in this file (`/32` and others).

hopkiw: The metadata server may return addresses with or without subnet masks depending on what is specified by the user, and for host local routes in some cases uses `/32` notation and in other cases does not. that's the revelation prompting this PR.

The only case where we need to modify anything provided by the metadata server is the `/32` case, since `1.1.1.1` and `1.1.1.1/32` are synonymous but won't be compared as synonymous since they aren't the same string. to handle this one case, we strip `/32` from all addresses prior to doing comparisons. 

`TestAddressDiff()` asks whether anything has changed from the previous instance of metadata to the current instance of metadata, checking the enablement logic, and doesn't seem relevant to this PR.

illfelder: > The metadata server may return addresses with or without subnet masks depending on what is specified by the user, and for host local routes in some cases uses /32 notation and in other cases does not. that's the revelation prompting this PR.

We should have tests where we parse the content of the metadata server with and without the subnet mask specified.

hopkiw: the metadata parsing isn't changing with this PR. please elaborate what test you think we need here

illfelder:
The code change in this pull request looks good to me. I still have concerns that we aren't adequately testing the regression which caused the issue you are attempting to solve. I propose that we check this in, but leave an open issue to improve unit test coverage.

Does this plan work for you @hopkiw?

illfelder:
Filed issue #14 

/lgtm

hopkiw:
> The code change in this pull request looks good to me. I still have concerns that we aren't adequately testing the regression which caused the issue you are attempting to solve. I propose that we check this in, but leave an open issue to improve unit test coverage.
> 
> Does this plan work for you @hopkiw?

sure, let's continue discussion in an issue and get this merged in for now

PR #46, opened 2020-05-27
Enable diagnostics tool by default for GA 

YangLu1031:
To use [diagnostics tool](go/vanadium/guest_os/guest_diagnostics_tool#run-gcloud-cli-to-collect-logs), Users need to 
1) enable `roles/iam.serviceAccountTokenCreator` for their project.
2) enable `enable-diagnostics` metadata flag.
3) trigger the gcloud command to start.

From PM, TL and customer voices, step 2 seems unneeded, we decided to remove the `enable-diagnostics` flag for customer experience and gain more usage for GA.

### DIFF
--- a/google_guest_agent/diagnostics.go
+++ b/google_guest_agent/diagnostics.go
@@ -28,7 +28,7 @@ const diagnosticsCmd = `C:\Program Files\Google\Compute Engine\diagnostics\diagn
 
 var (
 	diagnosticsRegKey   = "Diagnostics"
-	diagnosticsDisabled = true
+	diagnosticsDisabled = false
 )
 
 type diagnosticsEntry struct {
@@ -70,7 +70,7 @@ func (d *diagnosticsMgr) disabled(os string) (disabled bool) {
 		}
 	}()
 
-	// Diagnostics are opt-in and disabled by default.
+	// Diagnostics are opt-in and enabled by default.
 	var err error
 	var enabled bool
 	enabled, err = strconv.ParseBool(config.Section("diagnostics").Key("enable").String())
--- a/google_guest_agent/diagnostics_test.go
+++ b/google_guest_agent/diagnostics_test.go
@@ -46,7 +46,7 @@ func TestDiagnosticsDisabled(t *testing.T) {
 		md   *metadata
 		want bool
 	}{
-		{"not explicitly enabled", []byte(""), &metadata{}, true},
+		{"not explicitly enabled", []byte(""), &metadata{}, false},
 		{"enabled in cfg only", []byte("[diagnostics]\nenable=true"), &metadata{}, false},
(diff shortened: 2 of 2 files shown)


google_guest_agent/diagnostics_test.go
```diff
@@ -46,16 +46,16 @@ func TestDiagnosticsDisabled(t *testing.T) {
 		md   *metadata
 		want bool
 	}{
-		{"not explicitly enabled", []byte(""), &metadata{}, true},
+		{"not explicitly enabled", []byte(""), &metadata{}, false},
```
pjh, line 49: I don't really understand these tests (what is "data" in the struct above? What is "cfg" in the names below?), but I'm guessing that we need a lot fewer of these cases now. Why do we care about looking at instance metadata or project metadata at all now?

YangLu1031: "data" structure and "cfg" are from [instance_configs](https://cloud.google.com/compute/docs/instances/windows/creating-managing-windows-instances#configure-windows-features), which is used to control enabling or disabling windows instance features.
Thought to leave these test cases to verify no matter how metadata & instance configs are, the result is the same.
But right, this is UT not functional test, will remove them.

pjh: Thanks, this is better - it much more clearly tests the behavior that matters.

pjh:
LGTM

@adjackura PTAL when you have a chance, thanks.

google_guest_agent/diagnostics.go
```diff
@@ -60,32 +58,8 @@ func (d *diagnosticsMgr) timeout() bool {
 }
 
 func (d *diagnosticsMgr) disabled(os string) (disabled bool) {
```
adjackura, line 60: All features we have ways to disable, i dont think you should take out the metadata or config options.

pjh: Is there a reason that this should be disable-able, other than other features of the agent can be disabled?

It seems harmless to leave this enabled and having to activate it before using the gcloud command seems like an annoyance (and some customers agree, from PR description), but I'm no expert.

YangLu1031: How about taking out of the metadata options and leave the config options enabled by default?

YangLu1031: @adjackura Can you take a look at my latest change? The config options is there, so we can disable it.

adjackura: Making agent features disablable via metadata is pretty important for customers that want to restrict what users can instruct the agent to do. Enabled by default is fine, and very few users ever disable the features, but all other features have clear ways to disable so this should match those.

YangLu1031: Ok, leave all the config & metadata options, just enable it by default if not defined.

YangLu1031:
@adjackura Addressed the comments, PTAL, thanks!

Issue #54, opened 2020-06-05
Local route not set on MIG instances with ipv6 disabled behind TCP ILB


rsicart:
Thanks to GCP Support on a case, we found that the following setup causes guest-agent to not set local route to ILB IP address:

* 1 http healthcheck on port 80
* 1 instance template
* 1 MIG with 3 instances
* 1 TCP ILB
* 1 BackendService

Instances are debian-9 based, with a custom image build on top of debian-cloud images.
We use packer to build images, and to setup instance middlewares and parameters.

By default, we disable ipv6 with sysctl flags:

* `net.ipv6.conf.all.disable_ipv6=1`
* `net.ipv6.conf.default.disable_ipv6=1`

When instance boots, and launches guest-agent, we get this error:

```
GCEGuestAgent Error main.go:119: error running &main.addressMgr{} manager: Internet Systems Consortium
DHCP Client 4.3.5#012Copyright 2004-2016 Internet Systems Consortium.#012All rights reserved.#012For info, please visit https://www.isc.org/software/dhcp/#012#012no link-local IPv6 address for eth0#012#012If you think you have received this message due to a bug rather#012than a configuration issue please read the section on submitting#012bugs on either our web page at www.isc.org or in the README file#012before submitting a bug.  These pages explain the proper#012process and the information we find helpful for debugging..#012#012exiting.
```

That happens right before setting up local route to ILB ip address.

Thanks in advance.

ti-mo:
Just ran into the exact same issue right as this was posted! Adding and removing instances to/from a backend at runtime works fine, but rebooting (or recreating) it will cause the `google-guest-agent` to fail its startup sequence and skip fetching/adding these addresses.

rsicart:
As an patch solution, if you're blocked, you can set all flags to 0, or remove them:

* `net.ipv6.conf.all.disable_ipv6=0`
* `net.ipv6.conf.default.disable_ipv6=0`

And specific network interface, p.e.:

* `net.ipv6.conf.eth0.disable_ipv6=0`

And reboot the instance.

But I think that Guest Agent should handle the case where ipv6 is not enabled.

hopkiw:
what version of the guest agent are you using when you encounter this bug?

hopkiw:
based on the log output you give above, I believe this issue was fixed in #43 
can you update to version `20200529.00` or later, and see if the issue persists?

rsicart:
[...]

PR #55, opened 2020-06-08
Fix: do not configure IPv6 if disabled in sysctl

rsicart:
See Issue #54 for more details.

Just an example patch, I never used Golang before, so please, test and do not hesitate to rewrite the patch.
Thanks in advance.


rsicart:
@googlebot I signed it!

rsicart:
Issue fixed by in version `20200529.00` by bugfix https://github.com/GoogleCloudPlatform/guest-agent/pull/43 .

PR #57, opened 2020-06-15
Improve the logging of diagnostics

YangLu1031:
As per our discussions with customers, need to improve the output status / progress to the serial port the following, e.g.:
- $DATE-TIME$ $GCE AGENT$ ($VERSION$): export-logs requested
- $DATE-TIME$ $GCE AGENT$ ($VERSION$): export-logs collecting logs
- $DATE-TIME$ $GCE AGENT$ ($VERSION$): export-logs uploading to [[bucket URI]]
- $DATE-TIME$ $GCE AGENT$ ($VERSION$): export-logs completed

### DIFF
--- a/google_guest_agent/diagnostics.go
+++ b/google_guest_agent/diagnostics.go
@@ -89,6 +89,7 @@ func (d *diagnosticsMgr) disabled(os string) (disabled bool) {
 }
 
 func (d *diagnosticsMgr) set() error {
+	logger.Infof("Diagnostics: logs export requested.")
 	diagnosticsEntries, err := readRegMultiString(regKeyBase, diagnosticsRegKey)
 	if err != nil && err != errRegNotExist {
 		return err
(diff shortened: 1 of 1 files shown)


google_guest_agent/diagnostics.go
```diff
@@ -89,6 +89,7 @@ func (d *diagnosticsMgr) disabled(os string) (disabled bool) {
 }
 
 func (d *diagnosticsMgr) set() error {
+	logger.Infof("export-logs requested")
```
adjackura, line 92: Logs should be complete sentences, so capitalization and punctuation.

Also we should stick with the diagnostics naming as this is not a generic log exporter, or just not have this prefixed as no other manager has prefixed logs (although that's not a horrible idea)
so maybe:
"Diagnostics: logs export requested." or "Logs export requested."
"Diagnostics: collecting logs from the system:" or "Collecting logs from the system:"

YangLu1031: Done

YangLu1031:
@adjackura Addressed your comments, please take a look. Thanks!

PR #58, opened 2020-06-21
Reject the following diagnostics requests if an existing job is running

YangLu1031:
As the diagnostics log collection job consumes much resources, we decided to reject the following requests if an existing job is running, to make sure it won't impact the performance of vm.

### DIFF
--- a/google_guest_agent/diagnostics.go
+++ b/google_guest_agent/diagnostics.go
@@ -19,6 +19,7 @@ import (
 	"os/exec"
 	"reflect"
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
@@ -29,6 +30,9 @@ const diagnosticsCmd = `C:\Program Files\Google\Compute Engine\diagnostics\diagn
 var (
 	diagnosticsRegKey   = "Diagnostics"
 	diagnosticsDisabled = false
+	// Indicate whether an existing job is runing to collect logs
+	// 0 -> not running, 1 -> running
+	isDiagnosticsRunning int32 = 0
 )
 
 type diagnosticsEntry struct {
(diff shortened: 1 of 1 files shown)


google_guest_agent/diagnostics.go
```diff
@@ -116,7 +120,11 @@ func (d *diagnosticsMgr) set() error {
 	if entry.Trace {
 		args = append(args, "-trace")
 	}
-
+	// If no existing running job, set it to 1 and block other requests
+	if !atomic.CompareAndSwapInt32(&isDiagnosticsRunningInt32, 0, 1) {
```
ning-yang, line 124: maybe Andrew can give better suggestions(i am not an expert in go) but i think mutex is easy to use and read.

adjackura: A mutex would be easier to read, but you would lose the checking ability. You can also just use a channel which would be preferable:
https://play.golang.org/p/wiX0zeRSZnT

YangLu1031: I see. Found some discussions about Mutex vs Channel, [link1](https://github.com/golang/go/wiki/MutexOrChannel) & [link2](https://medium.com/mindorks/https-medium-com-yashishdua-synchronizing-states-using-mutex-vs-channel-in-go-25e646c83567). 
IIUC, mutex is preferable for this case, as we only need to change state without passing any data.
And I chose atomic is because it locks and update the gate variable in one function, it works like mutex but simpler.

ning-yang: i still think mutex is easy to read but if you want to use CompareAndSwapInt32, you need to use it in the thread as well. i think CompareAndSwapInt32 include memory fence which make sure all processor gets the correct value.

google_guest_agent/diagnostics.go
```diff
@@ -29,6 +30,9 @@ const diagnosticsCmd = `C:\Program Files\Google\Compute Engine\diagnostics\diagn
 var (
 	diagnosticsRegKey   = "Diagnostics"
 	diagnosticsDisabled = false
+	// Indicate whether an existing job is runing to collect logs
+	// 0 -> not running, 1 -> running
+	isDiagnosticsRunningInt32 int32 = 0
```
adjackura, line 35: just isDiagnosticsRunning, don't be redundant with the type

google_guest_agent/diagnostics.go
adjackura: CompareAndSwapInt32 would be the simplest solution here (least code), and really I'm fine with it as this is a fairly simple lock, but as pointed out you need to do this atomicly at line 137 as well (use SwapInt32 to avoid a broken state).

YangLu1031: Got it. Done

PR #63, opened 2020-07-15
remove distros check, disable NM according to network manager running status 

gaohannk:


google_guest_agent/addresses.go
```diff
@@ -531,6 +531,11 @@ func enableNetworkInterfaces() error {
 	}
 }
 
+func isNetworkManagerRunning() bool {
```
hopkiw, line 534: this is a one-command function, you can just return this logic inline since it's not used elsewhere and isn't suitable for a unit test.

secondly, this adds a dependency on systemd, and assumes 'systemctl' can be found in the path. are you intentionally trying to rely on runCmd returning an error in the case where that's not true? are you sure we only run networkmanager on systemd systems?

gaohannk: Oh. I see. For centos/rhel 6, I think it's sysvinit, there are ways to detect init system without by check distros.

google_guest_agent/addresses.go
```diff
@@ -532,8 +532,22 @@ func enableNetworkInterfaces() error {
 }
 
 func isNetworkManagerRunning() bool {
-	res := runCmd(exec.Command("systemctl", "is-active", "--quiet", "NetworkManager"))
-	return res == nil
+	// SystemD
+	if systemctl, err := exec.LookPath("systemctl"); err == nil {
+		return runCmd(exec.Command(systemctl, "is-active", "--quiet", "NetworkManager")) == nil
```
hopkiw, line 537: this assumes that if systemctl exists, that systemd is running. that's usually true but not necessarily. might want to switch this to checking for the existence of /var/run/systemd, which means systemd is running. or just look for running network manager process.

google_guest_agent/addresses.go
```diff
@@ -532,8 +532,22 @@ func enableNetworkInterfaces() error {
...
+	}
+	// Upstart
+	if runCmd(exec.Command("/sbin/init", "--version")) == nil {
```
hopkiw, line 540: all init systems provide an /sbin/init, but i don't know if all of them are safe to invoke directly like this on a running system

google_guest_agent/addresses.go
```diff
@@ -532,8 +532,22 @@ func enableNetworkInterfaces() error {
...
+		outString :=runCmdOutput(exec.Command("status", "network-manager")).out
+		return strings.Contains(outString, "running")
+	}
+	// Sysvinit
+	if _, err := os.Stat("/etc/init.d/network"); err == nil {
+		return runCmd(exec.Command("service  ", "network", "status")) == nil
```
hopkiw, line 546: i don't think asking for a 'network' service necessarily implies NetworkManager is running. is there maybe a mechanism NetworkManager itself provides? on my gLinux system, i see /var/run/NetworkManager exists, if we can confirm that is uniformly true, maybe it will simplify this logic?

google_guest_agent/addresses.go
```diff
@@ -503,7 +503,7 @@ func enableNetworkInterfaces() error {
 	switch {
 	case osRelease.os == "sles":
 		return enableSLESInterfaces(googleInterfaces)
-	case (osRelease.os == "rhel" || osRelease.os == "centos") && osRelease.version.major >= 7:
+	case isNetworkManagerRunning():
```
hopkiw, line 506: you're mixing different sorts of tests into the case. what if there was a sles system running network manager?

additionally, i think we only want to create this file on EL-type distros that will respect the /etc/sysconfig system that we modify with the `disableNM` function. if NetworkManager is running on Debian, I am not sure anything would need to be done. Will look into that.

google_guest_agent/addresses.go
gaohannk: The test condition I found is like this:
[[ `/sbin/init --version` =~ upstart ]] && echo yes || echo no
Since It's after testing in systemd, so I omit this string match logic.

google_guest_agent/addresses.go
gaohannk: i thought "network" and "networkManager" are refer to the same thing. But obviously that's not true. it should just named network-manager
/etc/init.d/network-manager

gaohannk:
discuss offline with @hopkiw, This change is not required.

PR #83, opened 2020-10-15
do not overwrite ifcfg files on SLES

jgleissner:
On SLES, the guest agent always writes an ifcfg file for any secondary NIC regardless whether there is one already. This makes it impossible to use a deviating configuration. This PR adds a check whether an ifcfg file for the interface that is being set up already exists and skips setup if that is the case.

### DIFF
--- a/google_guest_agent/addresses.go
+++ b/google_guest_agent/addresses.go
@@ -534,7 +534,12 @@ func enableSLESInterfaces(interfaces []string) error {
 	var priority = 10100
 	for _, iface := range interfaces {
 		var ifcfg *os.File
-		ifcfg, err = os.Create("/etc/sysconfig/network/ifcfg-" + iface)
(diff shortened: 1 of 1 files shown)


jgleissner:
@googlebot I signed it!

winterfreshx:
For review @hopkiw @zmarano

hopkiw:
Thank you for the PR. On GCE it is assumed that the guest agent is responsible for configuring secondary network interfaces. However, we do offer a way to opt out of letting Google manage the interfaces by editing the instance configuration file at /etc/default/instance_configs.cfg and setting:
```
[NetworkInterfaces]
setup = false
```
In which case management of secondary interfaces will be left up to the user. This existing option should support you, rather than making the somewhat complex new behavior change as proposed in this PR.

PR #87, opened 2020-10-27
Introduce struct to hold agent configuration values and print them on start

### LINKED ISSUE #85: Unable to inspect configuration of running google-guest-agent
Filed in [Ubuntu Bug #1901042](https://bugs.launchpad.net/ubuntu/+source/google-guest-agent/+bug/1901042) as well.

We have a cloud-images qualification test for the google-guest-agent to ensure the following daemons are enabled for GCE images:

- accounts_daemon
- clock_skew_daemon
- network_daemon

I could not find a way to inspect a currently running agent to determine the configuration in use. The agent only takes the following actions:

- run
- install
- remove
- start
- stop
- help

An additional action to list the configuration in use would be very helpful.

I also looked at the output of journalctl -u google-guest-agent.service, but the agent does not log configuration settings in the log, either.

davidkrauser:
This change introduces the AgentConfig struct. It consolidates configuration values into a single location, which makes it easier to inspect those values at runtime. We now print those configuration values on agent start.

This fixes #85.

This is a large change, so I broke it up into discrete commits that can be reviewed individually. Hopefully that makes it a bit easier :-)

### DIFF
--- a/google_guest_agent/accounts_test.go
+++ b/google_guest_agent/accounts_test.go
@@ -76,14 +76,18 @@ func TestAccountsDisabled(t *testing.T) {
 	for _, tt := range tests {
 		cfg, err := ini.InsensitiveLoad(tt.data)
 		if err != nil {
-			t.Errorf("test case %q: error parsing config: %v", tt.name, err)
+			t.Errorf("test case %q: error parsing ini data: %v", tt.name, err)
 			continue
 		}
 		if cfg == nil {
 			cfg = &ini.File{}
 		}
 		newMetadata = tt.md
-		config = cfg
+		config, err = parseIni(cfg)
+		if err != nil {
+			t.Errorf("test case %q: error parsing config: %v", tt.name, err)
+			continue
+		}
 		got := (&winAccountsMgr{}).disabled("windows")
 		if got != tt.want {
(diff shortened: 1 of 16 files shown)


davidkrauser:
@googlebot I signed it!

davidkrauser:
Hi @hopkiw.

I know previously you've stated that [you are not a fan of dumping the configuration to the log on start](https://github.com/GoogleCloudPlatform/guest-agent/issues/85#issuecomment-720728000). If I were to drop commit 366facf3b0831232568ea59edd1ceae282cd529b, what are your thoughts on the remaining changes as a general improvement?

I see the benefits as:
- Consolidation of default values to a single place makes it easy to see (in code) all of the defaults at a glance
- If a configuration value is used in more than one place, the default value only needs to be specified once
- Implicit behavior around the existence of some keys is made explicit (in code), which helps readability
- The additional structure helps enforce more uniform naming for new section/key additions to the config
- The additional structure makes it easier to do more complex operations with the config in the future if we chose to (config file content validation, etc.)

I see some of the downsides as:
- A bit more complexity
- Adding a new section/key to the config requires writing more plumbing
- Default values now live further from their actual usage
- This change is large, and could introduce bugs

Thanks,

David

PR #88, opened 2020-10-28
Only attempt to connect to snapshot service once

maxnelso:
If the snapshot service isn't available, this can consume a lot of resources attempting to reconnect.

### DIFF
--- a/google_guest_agent/snapshot_listener.go
+++ b/google_guest_agent/snapshot_listener.go
@@ -85,8 +85,8 @@ func listenForSnapshotRequests(address string, requestChan chan<- *sspb.GuestMes
 		logger.Infof("Attempting to connect to snapshot service at %s.", address)
 		conn, err := grpc.Dial(address, grpc.WithInsecure())
 		if err != nil {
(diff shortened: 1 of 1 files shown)


google_guest_agent/snapshot_listener.go
```diff
@@ -85,8 +85,8 @@ func listenForSnapshotRequests(address string, requestChan chan<- *sspb.GuestMes
 		logger.Infof("Attempting to connect to snapshot service at %s.", address)
 		conn, err := grpc.Dial(address, grpc.WithInsecure())
 		if err != nil {
-			logger.Errorf("Failed to connect: %v.", err)
-			continue
+			logger.Errorf("Failed to connect to snapshot service: %v.", err)
+			break
```
hopkiw, line 89: altho it works bc the function is entirely one for loop, i think the intention here is maybe better captured by 'return'?

maxnelso: Fixed (in both places)!

hopkiw:
ugh
/approve

PR #89, opened 2020-11-06
Add checks for if script is root owned/writable

maxnelso:
This is to address a possible privilege escalation security issue whereby an attacker could run their own script via the guest agent.


maxnelso:
A new and improved version of this is coming!

PR #91, opened 2020-12-07
Adding support for Windows user account password lengths between 15 and 255 characters.

bkatyl:
Adding support for Windows user account password lengths between 15 and 255 characters.

If no value is provided or if the value is less than 15, then 15 will be used for the password length.


google_guest_agent/windows_accounts.go
```diff
@@ -41,13 +41,20 @@ var (
 // requirements: https://technet.microsoft.com/en-us/library/cc786468.
 // Characters that are difficult for users to type on a command line (quotes,
 // non english characters) are not used.
-func newPwd() (string, error) {
+func newPwd(userPwLgth int) (string, error) {
 	pwLgth := 15
```
adjackura, line 45: Maybe rename to minPwLgth

google_guest_agent/windows_accounts.go
```diff
@@ -41,13 +41,20 @@ var (
...
+	maxPwLgth := 255
 	lower := []byte("abcdefghijklmnopqrstuvwxyz")
 	upper := []byte("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
 	numbers := []byte("0123456789")
 	special := []byte(`~!@#$%^&*_-+=|\(){}[]:;<>,.?/`)
 	chars := bytes.Join([][]byte{lower, upper, numbers, special}, nil)
+	if userPwLgth > maxPwLgth {
+		pwLgth = maxPwLgth
+	}
+	if userPwLgth > pwLgth {
```
adjackura, line 55: This logic is setting pwLgth to userPwLgth no matter if its greater than maxPwLgth.

google_guest_agent/windows_accounts.go
bkatyl: Done

google_guest_agent/windows_accounts.go
bkatyl: Great catch. I moved the user defined password length above the max length evaluation.

adjackura (approved):

PR #98, opened 2021-02-16
call restorecon on ssh host keys

hopkiw:


google_guest_agent/instance_setup.go
```diff
@@ -263,6 +263,7 @@ func generateSSHKeys() error {
 			logger.Warningf("Generated key is malformed, not uploading")
 		}
 	}
+	runCmd(exec.Command("restorecon", "-FR", hostKeyDir))
```
zmarano, line 266: Probably should check for its existence this like we do here: https://github.com/GoogleCloudPlatform/guest-agent/blob/a0da4b0f3d552272d53101ffb0d65151a07fcd50/google_guest_agent/oslogin.go#L308

hopkiw: the reverse, actually - we don't care whether it succeeded or failed, so we just 'try' blindly. path check can be removed there. see also restorecon invocation in `non_windows_accounts.go`

zmarano: Oh ok then lets make it consistent everywhere :)

PR #102, opened 2021-03-16
Enable creating Google users with GID.

dexter-1:
On COS, /home is stateful but /etc is stateless. So on reboots, all the
users are wiped but their home directories persist. A previous commit
was submitted to create a new user with the UID of their home directory
if it already exists. We need to also create the user with the GID of
their home directory, since it is not necessarily the case that the
UID == GID for a given user.

### DIFF
--- a/google_guest_agent/accounts_unix.go
+++ b/google_guest_agent/accounts_unix.go
@@ -23,19 +23,27 @@ import (
 	"syscall"
 )
 
-func getUID(path string) string {
+func getUIDAndGID(path string) (string, string) {
 	if dir, err := os.Stat(path); err == nil {
 		if stat, ok := dir.Sys().(*syscall.Stat_t); ok {
-			return fmt.Sprintf("%d", stat.Uid)
+			return fmt.Sprintf("%d", stat.Uid), fmt.Sprintf("%d", stat.Gid)
 		}
 	}
-	return ""
+	return "", ""
 }
 
-func createUser(username, uid string) error {
+func createUser(username, uid, gid string) error {
 	useradd := config.Section("Accounts").Key("useradd_cmd").MustString("useradd -m -s /bin/bash -p * {user}")
 	if uid != "" {
 		useradd = fmt.Sprintf("%s -u %s", useradd, uid)
+	}
+	if gid != "" {
+		groupadd := config.Section("Accounts").Key("groupadd_cmd").MustString("groupadd {group}")
+		groupadd = fmt.Sprintf("%s -g %s", groupadd, gid)
+		if err := runCmd(createUserGroupCmd(groupadd, "", username)); err != nil {
+			return err
+		}
+		useradd = fmt.Sprintf("%s -g %s", useradd, gid)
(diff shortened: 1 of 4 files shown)


google_guest_agent/non_windows_accounts.go
```diff
@@ -345,12 +345,12 @@ func createUserGroupCmd(cmd, user, group string) *exec.Cmd {
 // createGoogleUser creates a Google managed user account if needed and adds it
 // to the configured groups.
 func createGoogleUser(user string) error {
-	var uid string
+	var uid, gid string
 	if config.Section("Accounts").Key("reuse_homedir").MustBool(false) {
-		uid = getUID(fmt.Sprintf("/home/%s", user))
+		uid, gid = getUIDAndGID(fmt.Sprintf("/home/%s", user))
```
dorileo, line 350: Given that the `useradd_cmd` can be configured, the user may have changed it to pass in the `-d` argument and the home directory will be set to somewhere other than `/home`. We have a couple of options to avoid it happening. I'll not block this PR because of it since the behavior was introduced by another patch and got the behavior integrated already. Here's my ask: Please add a TODO comment around the path derivation in this function stating the problem I described.

google_guest_agent/accounts_unix.go
```diff
@@ -23,20 +23,28 @@ import (
...
+	return "", ""
 }
 
-func createUser(username, uid string) error {
+func createUser(username, uid, gid string) error {
 	useradd := config.Section("Accounts").Key("useradd_cmd").MustString("useradd -m -s /bin/bash -p * {user}")
 	if uid != "" {
 		useradd = fmt.Sprintf("%s -u %s", useradd, uid)
 	}
+	if gid != "" {
```
dorileo, line 40: You want to first add the group then the user, don't you? If we create a user first its group will be also created and then if you later to add the group the group will already exist resulting into a error i.e. `groupadd: group '<user>' already exists`.

dorileo (changes requested):

google_guest_agent/accounts_unix.go
dorileo: I misread the change in the first pass. It's indeed creating the group first.

dorileo: I'd rewrite this code to something like this:

```golang
func createUser(username, uid string, gid string) error {
       if gid != "" {
		groupadd := config.Section("Accounts").Key("groupadd_cmd").MustString("groupadd {group}")
		groupadd = fmt.Sprintf("%s -g %s", groupadd, gid)
		if err := runCmd(createUserGroupCmd(groupadd, "", username)); err != nil {
			return err
		}
       }

        useradd := config.Section("Accounts").Key("useradd_cmd").MustString("useradd -m -s /bin/bash -p * {user}")
        if uid != "" {
                useradd = fmt.Sprintf("%s -u %s", useradd, uid)
        }
        return runCmd(createUserGroupCmd(useradd, username, ""))
}
```

PR #106, opened 2021-04-13
start sshd

hopkiw:
* start sshd explicitly if not started
* remove non-systemd service handling

### DIFF
--- a/google_guest_agent/oslogin.go
+++ b/google_guest_agent/oslogin.go
@@ -103,9 +103,19 @@ func (o *osloginMgr) set() error {
 		logger.Errorf("Error updating group.conf: %v.", err)
 	}
 
-	for _, svc := range []string{"ssh", "sshd", "nscd", "unscd", "systemd-logind", "cron", "crond"} {
+	for _, svc := range []string{"nscd", "unscd", "systemd-logind", "cron", "crond"} {
 		if err := restartService(svc); err != nil {
 			logger.Errorf("Error restarting service: %v.", err)
+		}
+	}
+
+	// SSH should be explicitly started if not running.
+	for _, svc := range []string{"ssh", "sshd"} {
+		if err := startService(svc, true); err != nil {
+			logger.Errorf("Error restarting service: %v.", err)
+		} else {
+			// Stop on first matching, to avoid double restarting.
+			break
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


google_guest_agent/oslogin.go
```diff
@@ -103,12 +103,22 @@ func (o *osloginMgr) set() error {
 		logger.Errorf("Error updating group.conf: %v.", err)
 	}
 
-	for _, svc := range []string{"ssh", "sshd", "nscd", "unscd", "systemd-logind", "cron", "crond"} {
+	for _, svc := range []string{"nscd", "unscd", "systemd-logind", "cron", "crond"} {
```
zmarano, line 106: We have to restart sshd if oslogin is disabled (sshd configs change). So we still need to keep this here as well right?

hopkiw: ssh is still being restarted, it was only moved down below this

google_guest_agent/oslogin.go
```diff
@@ -103,12 +103,22 @@ func (o *osloginMgr) set() error {
...
 		if err := restartService(svc); err != nil {
 			logger.Errorf("Error restarting service: %v.", err)
 		}
 	}
 
+	// SSH should be explicitly started if not running.
+	for _, svc := range []string{"ssh", "sshd"} {
+		if err := startService(svc, true); err != nil {
```
hopkiw, line 114: here. true argument means 'restart it if already running'

zmarano: Oh I see, why have the restartService function then? Effectively startService now does both.

hopkiw: they map to systemctl behaviors, restartService only restarts if already running.

google_guest_agent/oslogin.go
```diff
@@ -340,35 +350,28 @@ func createOSLoginSudoersFile() error {
...
+			return runCmd(exec.Command(systemctl, "start", servicename+".service"))
 		}
-		return nil
-	}
-	initService := "/etc/init.d/" + servicename
-	if _, err := os.Stat(initService); err == nil {
-		if err := runCmd(exec.Command(initService, "status")); err == nil {
-			return runCmd(exec.Command(initService, "restart"))
+		if started && restart {
+			return runCmd(exec.Command(systemctl, "restart", servicename+".service"))
```
zmarano, line 373: Why not just call restartService(servicename)- at least this would make a bit more sense to me than dup'ing the exec call.

hopkiw: that would dupe the is-active call. i thought this was more readable and mapped to logical behaviors.

PR #126, opened 2021-09-07
refactor

gaohannk:


gaohannk:
We don't want to introduce interface just for mocking and testing.

PR #128, opened 2021-09-08
Integration test

gaohannk:


google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -4,9 +4,49 @@ package main
...
+		t.Errorf("failed creating test user")
+	}
+	if exist, err := userExists(testUser); exist != true || err != nil {
+		t.Errorf("tesr user %s should exist", testUser)
+	}
+	cmd := exec.Command("groups", testUser)
+	out := runCmdOutput(cmd).Stdout()
+	groups := strings.Split(strings.TrimSpace(strings.Split(out, ":")[1]), " ")
+	for _, group := range groups {
+		if !strings.Contains("adm,dip,docker,lxd,plugdev,video,google-sudoers", group) {
```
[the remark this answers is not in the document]
gaohannk: I understand this is not a safe way. I am lazy here as golang don't have set. I think we need check on both direction. i.e. these group has the target user and the target user only added in these groups.

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -4,9 +4,49 @@ package main
...
+			t.Errorf("test user has been added to an unexpected group")
+		}
+	}
+	if err := createGoogleUser(testUser); err == nil {
+		t.Errorf("expected user exist and return error but not")
+	}
+}
+
+func TestRemoveGoogleUser(t *testing.T) {
```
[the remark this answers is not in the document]
gaohannk: Yes, I think we should also check the home directory and AuthorizedKeysFile

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -4,9 +4,49 @@ package main
...
+	cmd := exec.Command("groups", testUser)
+	out := runCmdOutput(cmd).Stdout()
+	groups := strings.Split(strings.TrimSpace(strings.Split(out, ":")[1]), " ")
+	for _, group := range groups {
+		if !strings.Contains("adm,dip,docker,lxd,plugdev,video,google-sudoers", group) {
+			t.Errorf("test user has been added to an unexpected group")
+		}
+	}
+	if err := createGoogleUser(testUser); err == nil {
+		t.Errorf("expected user exist and return error but not")
```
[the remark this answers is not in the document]
gaohannk: Here we want to make sure the second create operation return error as expected.

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -4,9 +4,49 @@ package main
...
+}
+
+func TestRemoveGoogleUser(t *testing.T) {
+	if err := createGoogleUser(testUser); err != nil {
+		t.Errorf("failed creating test user")
+	}
+	if err := removeGoogleUser(testUser); err != nil {
+		t.Errorf("failed when remove google user")
+	}
+	cmd := exec.Command("groups", testUser)
```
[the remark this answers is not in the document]
gaohannk: Good catch. Just aware that remove user will also remove it in groups.

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -4,9 +4,49 @@ package main
...
+	ret := runCmdOutput(cmd)
+	if ret.ExitCode() != 1 {
+		t.Errorf("expected groups command return error code 1, but actually error is %d, %s", ret.ExitCode(), ret.Stderr())
+	}
+	if err := removeGoogleUser(testUser); err == nil {
```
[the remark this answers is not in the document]
gaohannk: We want to make sure remove an removed user should return error

google_guest_agent/accounts_unix.go
```diff
@@ -12,6 +12,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
+//go:build !windows
```
hopkiw, line 15: you can go ahead and remove these - we've fixed the presubmit to target our current golang version

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
 )
 
+const (
+	testUser       = "integration-test-user"
+	expectedGroups = []string{"adm", "dip", "docker", "lxd", "plugdev", "video", "google-sudoers"}
+)
+
+func TestCreateGoogleUser(t *testing.T) {
+	if exist, err := userExists(testUser); exist == true {
+		t.Errorf("test user should not exist")
```
hopkiw, line 21: make this t.Fatalf since you don't do anything to handle it

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+	}
+	if err := createGoogleUser(testUser); err != nil {
+		t.Errorf("failed creating test user")
+	}
+	if exist, err := userExists(testUser); exist != true || err != nil {
+		t.Errorf("tesr user should exist")
```
hopkiw, line 27:
```suggestion
		t.Errorf("test user should exist")
```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+	}
+	cmd := exec.Command("groups", testUser)
+	ret := runCmdOutput(cmd)
+	if ret.ExitCode() != 0 {
+		t.Errorf("test user should be added to group")
```
hopkiw, line 32:
```suggestion
		t.Errorf("failed looking up groups for user: stdout:%s stderr:%s", ret.Stdout(), ret.Stderr())
```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+	"fmt"
+	"os"
 	"os/exec"
+	"strings"
 	"testing"
 )
 
+const (
+	testUser       = "integration-test-user"
+	expectedGroups = []string{"adm", "dip", "docker", "lxd", "plugdev", "video", "google-sudoers"}
```
hopkiw, line 16: this is a configuration value, not a constant. you can set a list here IF you also set it in the config global, otherwise suggest  populating this at runtime with the same code used in `createGoogleUser()`:
```
const (
    defaultgroupstring := "adm,dip,docker,lxd,plugdev,video"
)
[ ... later in the test function ... ]
expectedGroups := config.Section("Accounts").Key("groups").MustString(defaultgroupstring)
```

we may also want a test that explicitly sets this to something else, and validates that it gets used during user creation

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+	for _, group := range groups {
+		if !contains(group, expectedGroups) {
+			t.Errorf("test user has been added to an unexpected group %s", group)
+		}
+	}
+	for _, expected := range expectedGroups {
+		if !contains(expected, groups) {
+			t.Errorf("test user has not been added to group %s", expected)
+		}
+	}
```
hopkiw, line 44: think we should not do this check. the list of supplementary groups is what we will *try* to add the user to, but the group itself may not exist. we don't guarantee that every group listed in `[Accounts].groups` will be added because we don't create the groups.
```suggestion
```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+	if _, err := os.Stat(fmt.Sprintf("/home/%s", testUser)); err != nil {
+		t.Errorf("test user home directory is not exist")
+	}
+	if err := createGoogleUser(testUser); err == nil {
+		t.Errorf("user should exist and return error but not")
```
hopkiw, line 49:
```suggestion
		t.Errorf("createGoogleUser did not return error when creating user that already exists")
```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+	}
+}
+
+func TestRemoveGoogleUser(t *testing.T) {
+	if exist, err := userExists(testUser); exist == true {
```
hopkiw, line 54: go unit tests run in parallel. how will you prevent these tests from interfering with each other?

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+		if !contains(expected, groups) {
+			t.Errorf("test user has not been added to group %s", expected)
+		}
+	}
+	if _, err := os.Stat(fmt.Sprintf("/home/%s", testUser)); err != nil {
+		t.Errorf("test user home directory is not exist")
+	}
+	if err := createGoogleUser(testUser); err == nil {
+		t.Errorf("user should exist and return error but not")
+	}
```
hopkiw, line 50: you don't delete the user at the end, meaning this test will have an effect on the system under test - tests will now succeed or fail depending on their timing and ordering

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+		t.Errorf("failed when remove google user")
+	}
+	if exist, err := userExists(testUser); exist == true {
+		t.Errorf("test user should not exist")
+	}
+	if _, err := os.Stat(fmt.Sprintf("/home/%s", testUser)); err == nil {
+		t.Errorf("test user home directory should not exist")
+	}
+	if err := removeGoogleUser(testUser); err == nil {
+		t.Errorf("user has been removed and should return error but not")
```
hopkiw, line 70:
```suggestion
		t.Errorf("removeGoogleUser did not return error when removing user that doesn't exist")

```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -1,12 +1,76 @@
...
+	}
+	if err := removeGoogleUser(testUser); err != nil {
+		t.Errorf("failed when remove google user")
+	}
+	if exist, err := userExists(testUser); exist == true {
+		t.Errorf("test user should not exist")
+	}
+	if _, err := os.Stat(fmt.Sprintf("/home/%s", testUser)); err == nil {
+		t.Errorf("test user home directory should not exist")
+	}
```
hopkiw, line 68: we only remove the home directory if configured to do so. for this simple test, just check whether we were supposed to. but we need to add a test that explicitly sets this config option and confirms it took effect. that could be a different unit test, or just turn this test into one that deals with multiple cases
```suggestion
	if config.Section("Accounts").Key("deprovision_remove").MustBool(false) {
		if _, err := os.Stat(fmt.Sprintf("/home/%s", testUser)); err == nil {
			t.Errorf("test user home directory should not exist")
		}
	}
```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -3,10 +3,62 @@
...
 	"testing"
 )
 
+const (
+	testUser           = "integration-test-user"
+	defaultgroupstring = "adm,dip,docker,lxd,plugdev,video,google-sudoers"
+)
+
+func TestCreateAndRemoveGoogleUser(t *testing.T) {
+	if exist, err := userExists(testUser); exist == true {
```
hopkiw, line 19:
```suggestion
	if exist, err := userExists(testUser); exist {
```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -3,10 +3,62 @@
...
+	}
+	cmd := exec.Command("groups", testUser)
+	ret := runCmdOutput(cmd)
+	if ret.ExitCode() != 0 {
+		t.Errorf("failed looking up groups for user: stdout:%s stderr:%s", ret.Stdout(), ret.Stderr())
+	}
+	groups := strings.Split(strings.TrimSpace(strings.Split(ret.Stdout(), ":")[1]), " ")
+	expectedGroups := config.Section("Accounts").Key("groups").MustString(defaultgroupstring)
+	for _, group := range groups {
+		if !contains(group, strings.Split(expectedGroups, ",")) {
```
hopkiw, line 36: you're unnecessarily calling split on every iteration. try this:
```suggestion
	expectedGroupString := config.Section("Accounts").Key("groups").MustString(defaultgroupstring)
	expectedGroups := strings.Split(expectedGroupString, ",")
	for _, group := range groups {
		if !contains(group, expectedGroups) {
```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -3,10 +3,62 @@
...
+			t.Errorf("test user has been added to an unexpected group %s", group)
+		}
+	}
+	if _, err := os.Stat(fmt.Sprintf("/home/%s", testUser)); err != nil {
+		t.Errorf("test user home directory is not exist")
```
hopkiw, line 41:
```suggestion
		t.Errorf("test user home directory does not exist")
```

google_guest_agent/non_windows_accounts_integ_test.go
```diff
@@ -3,10 +3,62 @@
...
+		t.Errorf("removeGoogleUser did not remove user")
+	}
+	if exist, err := userExists(testUser); exist == true {
+		t.Errorf("test user should not exist")
+	}
+	if config.Section("Accounts").Key("deprovision_remove").MustBool(false) {
+		if _, err := os.Stat(fmt.Sprintf("/home/%s", testUser)); err == nil {
+			t.Errorf("test user home directory should not exist")
+		}
+	}
```
hopkiw, line 56: this won't work unless you initialize the config global

alternatively since we know the execution environment, you can remove this entire section and just test this in the config-manipulating tests we have to add next

```suggestion
```

hopkiw (changes requested):

PR #129, opened 2021-09-08
rename createUserGroupCmd to createOrDeleteUserGroupCmd

gaohannk:


google_guest_agent/non_windows_accounts.go
```diff
@@ -349,7 +349,7 @@ func removeExpiredKeys(keys []string) []string {
 
 // Replaces {user} or {group} in command string. Supports legacy python-era
 // user command overrides.
-func createUserGroupCmd(cmd, user, group string) *exec.Cmd {
+func createOrDeleteUserGroupCmd(cmd, user, group string) *exec.Cmd {
```
hopkiw, line 352: the name `createUserGroupCmd` means create a command, not create a user. it returns a command. if we really want to rename this, we could consider just `userGroupCmd` or `craftUserGroupCmd` or `newUserGroupCmd` or something

hopkiw (changes requested):

PR #130, opened 2021-09-09
Integration test: add windows account test

gaohannk:


google_guest_agent/accounts_unix.go
```diff
@@ -12,7 +12,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-// +build !windows
+//go:build !windows
```
hopkiw, line 15: restore this

hopkiw (commented):
well the main issue here is that we don't run integ tests for windows. so i don't think we should include this yet, knowing it would never be run. suggest moving to draft or closing for now.

gaohannk:
Close as no need now. Once we have windows integration test setup, reopen

PR #131, opened 2021-09-13
Integration Test: cloud skew manager

gaohannk:


google_guest_agent/accounts_unix.go
```diff
@@ -12,7 +12,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-// +build !windows
+//go:build !windows
```
hopkiw, line 15: restore these

google_guest_agent/cloud_integ_test.go
```diff
@@ -0,0 +1,12 @@
+//go:build integration
+
+package main
+
+import "testing"
+
+func TestCloudSkewManager(t *testing.T) {
+	var c = clockskewMgr{}
+	if err := c.set(); err != nil {
+		t.Errorf("failed to run cloud skew manager")
```
hopkiw, line 10: no check for how it actually works?

PR #132, opened 2021-09-13
Integration test: test getRelase  and test diagnose manager

gaohannk:


google_guest_agent/accounts_unix.go
```diff
@@ -12,6 +12,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
+//go:build !windows
```
hopkiw, line 15: restore these

google_guest_agent/diagnostics_integ_test.go
```diff
@@ -0,0 +1,15 @@
...
+package main
+
+import (
+	"testing"
+)
+
+func TestDiagnosticsManager(t *testing.T) {
+	var d = diagnosticsMgr{}
+	if !d.disabled("linux") {
+		t.Errorf("linux system does not support diagnose")
```
hopkiw, line 13: all we do is validate the disabled function? i don't think that's adding much value

google_guest_agent/getrelease_integ_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"testing"
+)
+
+func TestGetRelease(t *testing.T) {
+	osrelease := getRelease()
+	if osrelease.os == "" {
+		t.Errorf("failed to get os name")
+	}
+	if osrelease.version.String() == "" {
+		t.Errorf("failed to get os version")
```
hopkiw, line 16: any idea for checking that it worked successfully?

hopkiw (changes requested):
doesn't seem like these tests are ready for review

google_guest_agent/diagnostics_integ_test.go
gaohannk: i agree that this test is probably needed for linux as diagnosticsManager is only available on windows

google_guest_agent/getrelease_integ_test.go
gaohannk: My idea is that we could compare the what get release returned with metadata server. I didn't

PR #133, opened 2021-09-20
enable ipv6 on secondary interfaces

hopkiw:
Today we start 1 dhclient process for all secondary NICs as in `dhclient eth1 eth2 ... ethN`. This PR adds a second dhclient process for IPv6 as in `dhclient -6 eth1 eth2 ... ethN`.

We cannot use one process as IPv6 and IPv4 modes are mutually exclusive in isc-dhclient:
```
OPTIONS
       -4     Use the DHCPv4 protocol to obtain an IPv4 address and configuration parameters.  This is the default and cannot be combined with -6.
```

As of this change, Debian 10 VM with 2 NICs and the second NIC is on an IPv6 network:
```
root       326     1  0 Sep10 ?        00:00:00 /sbin/dhclient -4 -v -i -pf /run/dhclient.ens4.pid -lf /var/lib/dhcp/dhclient.ens4.leases -I -df /var/lib/dhcp/dhclient6.ens4.leases ens4
root       327     1  0 Sep10 ?        00:00:00 /sbin/dhclient -4 -v -i -pf /run/dhclient.ens5.pid -lf /var/lib/dhcp/dhclient.ens5.leases -I -df /var/lib/dhcp/dhclient6.ens5.leases ens5
root     12241     1  0 22:35 ?        00:00:00 dhclient ens5
root     12324 12171  0 22:35 ?        00:00:00 dhclient -6 ens5
root     12325 12324  0 22:35 ?        00:00:00 dhclient -6 ens5
```

### DIFF
--- a/google_guest_agent/addresses.go
+++ b/google_guest_agent/addresses.go
@@ -482,7 +482,8 @@ func configureIPv6() error {
 	return nil
 }
 
-// enableNetworkInterfaces runs `dhclient eth1 eth2 ... ethN`.
+// enableNetworkInterfaces runs `dhclient eth1 eth2 ... ethN`
+// and `dhclient -6 eth1 eth2 ... ethN`.
 // On RHEL7, it also calls disableNM for each interface.
 // On SLES, it calls enableSLESInterfaces instead of dhclient.
 func enableNetworkInterfaces() error {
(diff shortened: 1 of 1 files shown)


hopkiw:
there's a problem with the presubmit, filed a PR to fix.

zmarano:
/lgtm

However, I want to make sure we understand the use cases correctly first.

PR #144, opened 2021-11-03
A default umask of 0022 allows all users to read the contents of others home directories

irab:
Without setting `umask=0077` user credentials, such as gcloud tokens*, can be exposed to everyone logging into a VM. A umask of 0077 will set home directories to be only readable to the user owns the home directory.

Using this agent on RHEL, Fedora or [Ubuntu from 21.04](https://www.phoronix.com/scan.php?page=news_item&px=Ubuntu-21.04-Private-Home) will revert security controls that users expect to be present.

*I mean tokens written out by gcloud logs, as they're set to DEBUG by default, which logs auth tokens as in API responses.

### DIFF
--- a/google_guest_agent/oslogin.go
+++ b/google_guest_agent/oslogin.go
@@ -276,13 +276,13 @@ func updatePAMsshd(pamsshd string, enable, twofactor bool) string {
 	authGroup := "auth       [default=ignore] pam_group.so"
 	accountOSLogin := "account    [success=ok ignore=ignore default=die] pam_oslogin_login.so"
 	accountOSLoginAdmin := "account    [success=ok default=ignore] pam_oslogin_admin.so"
-	sessionHomeDir := "session    [success=ok default=ignore] pam_mkhomedir.so"
+	sessionHomeDir := "session    [success=ok default=ignore] pam_mkhomedir.so umask=0077"
 
 	if runtime.GOOS == "freebsd" {
 		authOSLogin = "auth       optional pam_oslogin_login.so"
 		authGroup = "auth       optional pam_group.so"
 		accountOSLogin = "account    requisite pam_oslogin_login.so"
-		accountOSLoginAdmin = "account    optional pam_oslogin_admin.so"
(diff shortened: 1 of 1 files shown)


simon-anz:
It's crazy that users can read other user homedir by default. That's like multi-user 101.

hopkiw:
We have had this discussion a few times in the past, and the conclusion was that we use the upstream behavior for this intentionally. The good news is that newer versions of pam_mkhomedir upstream now read the umask file from the system configuration, so this issue should self-resolve for you in time as these changes make their way into distributions.

irab:
@hopkiw Would you be open to a PR that added the umask as a setting in [instance_configs.cfg](https://github.com/GoogleCloudPlatform/guest-agent/blob/cfc05e2e8fbbf36ae22544299257318f65162b27/instance_configs.cfg)? 

Also just noticed that you now have an option to set `useradd_cmd` as well, which supports having a home directory UMASK setting in /etc/login.defs? Which process takes precedence in the user setup flow for home directory setup, the `useradd_cmd` [here](https://github.com/GoogleCloudPlatform/guest-agent/blob/cfc05e2e8fbbf36ae22544299257318f65162b27/google_guest_agent/accounts_unix.go#L35) or the pam sshd setup in this PR?

irab:
FYI - the linux-pam release with this functionality is [version 1.5.2](https://github.com/linux-pam/linux-pam/releases/tag/v1.5.2)

sohrab-:
> Also just noticed that you now have an option to set `useradd_cmd` as well, which supports having a home directory UMASK setting in /etc/login.defs? Which process takes precedence in the user setup flow for home directory setup, the `useradd_cmd` [here](https://github.com/GoogleCloudPlatform/guest-agent/blob/cfc05e2e8fbbf36ae22544299257318f65162b27/google_guest_agent/accounts_unix.go#L35) or the pam sshd setup in this PR?

@irab Those are two different managers. `instance_configs.cfg` is used by the account manager. The account manager is used when OS Login manager is disabled, i.e. only local users. When OS Login Manager is enabled those commands don't get triggered and everything is delegated to PAM.

PR #152, opened 2022-01-10
Allow disabling parts of the agent that imperatively configure the system

abbradar:
On NixOS we configure the whole system via our declarative configuration system. `/etc` is mostly immutable, so Guest Agent produces several error messages. These patches allow to (optionally) disable parts of Guest Agent that conflict with our configuration.


abbradar:
@NinjaTrappeur would you mind signing CLA for this?

hopkiw:
Can you open an issue to discuss this? We are intentionally not extending the instance configs concept.

abbradar:
Sure, I'll return to this tomorrow -- it's late in my TZ.

PR #156, opened 2022-02-17
Initial commit for AKC for handoff

bkatyl:


hopkiw:
/hold

can we discuss this offline. we already have an authorized keys command for linux, i would like to see if we can get all platforms on the same thing rather than starting from scratch.

google_authorized_keys/main.go
```diff
@@ -0,0 +1,137 @@
...
+
+func main() {
+	ctx := context.Background()
+	username := os.Args[1]
+
+	opts := logger.LogOpts{
+		LoggerName:     programName,
+		FormatFunction: logFormat,
+	}
+	opts.Writers = []io.Writer{&serialPort{"COM1"}, os.Stdout}
```
hopkiw, line 125: why would the akc print to the serial console?

google_authorized_keys/main.go
```diff
@@ -0,0 +1,137 @@
...
+	logger.Init(ctx, opts)
+	logger.Infof("Starting %s version %s for user %s.", programName, version, username)
+
+	projectID, err := getMetadataKey("/project/project-id")
```
hopkiw, line 129: is this a demo? where's the ssh key handling? are you going to use any code from the guest agent, here?

PR #165, opened 2022-05-13
Support .exe as an option for scripts and refactor runScript

patelne:
Adds support for windows binaries as part of the startup script.

Also refactors and adds some unit tests to make the code easier to understand.

### DIFF
--- a/google_metadata_script_runner/main_test.go
+++ b/google_metadata_script_runner/main_test.go
@@ -182,9 +182,9 @@ func TestNormalizeTmpFileForWindows(t *testing.T) {
 	tmpFilePath := "C:/Temp/file"
 
 	testCases := []struct {
-		metadataKey string
-		gcsScriptUrlPath string
-		want string
+		metadataKey      string
+		gcsScriptURLPath string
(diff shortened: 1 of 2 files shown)


bkatyl:
As we talked about in chat last week, I think it would be helpful to breakup this function so that the logic in the function is covered by unit tests. The exec calls are likely not testable, which is why I suggest breaking it out into 2 functions.

google_metadata_script_runner/main_test.go
```diff
@@ -178,7 +178,7 @@ func TestGetMetadata(t *testing.T) {
 	}
 }
 
-func TestNormalizeTmpFileForWindows(t *testing.T) {
+func TestNormalizeFilePathForWindows(t *testing.T) {
 	tmpFilePath := "C:/Temp/file"
 
 	testCases := []struct {
```
bkatyl, line 184: go/go-style/best-practices#t-field-labels

Can you please add the field labels to make the test more readable?

bkatyl (approved):

PR #167, opened 2022-05-26
Add authorized keys command to the Windows agent package.

wrigri:


packaging/googet/windows_agent_build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Should be run from the guest-agent directory.
+# Usage:
+#   ./packaging/googet/windows_agent_build.sh <version>
+
+version=$1
+
+GOOS=windows /tmp/go/bin/go build -ldflags "-X main.version=$version" -mod=readonly -o GCEWindowsAgent.exe ./google_guest_agent
+GOOS=windows /tmp/go/bin/go build -mod=readonly -o GCEAuthorizedKeysCommand.exe ./google_authorized_keys
```
bkatyl, line 10: I'd suggest using the same build flags in the same order. Also for the GCEAuthorizedKeysCommand.exe version, I think it's reasonable to use the same version as the agent. The version helps ensure we can identify if it's the correct/incorrect version of the binary.

wrigri: Done. Added the same version string to GCEAuthorizedKeysCommand.

hopkiw:
Presubmit fixed in https://github.com/GoogleCloudPlatform/guest-test-infra/pull/477

PR #168, opened 2022-06-01
no authorized keys on debian

hopkiw:
debian's build system autodetects and includes binaries; we aren't ready to ship this binary on linux, yet.


hopkiw:
ok first pass didn't work, i think that's not the right path at that moment in the build

hopkiw:
/unhold

tested in build workflow

PR #171, opened 2022-06-17
update for golang 1.16

hopkiw:
* update debian build rules to specify targets
* reset dependencies for 1.16


dntczdx:
Looks like it failed "goo" in prow/presubmit/packagebuild.

PR #172, opened 2022-06-22
try restoring module mode

hopkiw:
for some reason on ARM64 this was still needed


hopkiw:
/hold

ok why did this work with daisy but not with packagebuild presubmit

Issue #157, opened 2022-02-17
Bug determining expired keys in accountsMgr.diff()


nickkortman:
[This block](https://github.com/GoogleCloudPlatform/guest-agent/blob/main/google_guest_agent/non_windows_accounts.go#:~:text=%7D-,//%20If%20any%20on%2Ddisk%20keys%20have%20expired.,%7D,-//%20If%20we%27ve%20just) uses `getUserKeys` to determine if there are any keys that have expired since the last iteration. However, the keys in `sshKeys` are not of the format that `getUserKeys` expects (no "user:" prefix), so they always hit [this block](https://github.com/GoogleCloudPlatform/guest-agent/blob/main/google_guest_agent/non_windows_accounts.go#:~:text=idx%20%3A%3D%20strings.Index,%7D), which means that `getUserKeys` returns an empty list, and `diff` always returns true.

hopkiw:
Thanks for catching this, Nick. You're right, we're unnecessarily reconciling users' ssh keys every minute. Luckily, this is the good kind of failure mode, compared to not refreshing them :)

We are considering replacing this logic with an SSH authorized keys command, as we do with OS Login. We will either fix this bug or make that transition. Will leave this issue open for tracking or any discussion.

rogerhub:
Hi, can we have this fixed? It's causing confusing log spam for me:

```
2022-07-17 19:20:28.202 PDT non_windows_accounts.go:199] Invalid ssh key entry - unrecognized format: ssh-rsa ...
```

Without context, this message suggests that the guest agent might not be updating SSH keys properly, which is alarming. It looks like this bug was introduced in #142? @gaohannk Any thoughts about this?

Thanks!

hopkiw:
that message does suggest your agent is not updating keys properly. i don't believe it is related to this issue.

rogerhub:
The same underlying bug causes both the unnecessary refresh as well as the log spam.

I work on GCE by the way :)

PR #142, opened 2021-10-28
add malformed ssh key unit test 

gaohannk:
This test validate that malformed ssh key handled by guest-agent correctly. 
Two cases: 
- `sshkeys` without a colon 
- empty name `:sshkeys`
- refactor removeNonExpiredKeys

PR #175, opened 2022-08-19
correct expired key handling

hopkiw:
Refactor in #142 used getUserKeys in place of the old removeExpiredKeys, which was not valid. This PR restores the original logic.

Fixes: #157

### DIFF
--- a/google_guest_agent/non_windows_accounts.go
+++ b/google_guest_agent/non_windows_accounts.go
@@ -56,6 +56,16 @@ func compareStringSlice(first, second []string) bool {
 	return true
 }
 
+func removeExpiredKeys(keys []string) []string {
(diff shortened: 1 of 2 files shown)


koln67 (approved):
lgtm

When I ran this on the GCE VM, the bug from issue #157 did not occur.

ccjon:
I don't know how to determine when this might finally be resolved in the deployment pipeline. I've running ubuntu 18.04 vm's and they all still have this release:
google-guest-agent/bionic-updates,now 20220622.00-0ubuntu2~18.04.0 amd64
And iiic this fix is in 20220927.00 
Not a huge issue for me but I do monitor syslog ERROR reporting, so this is crying wolf every day

PR #176, opened 2022-08-22
Expired key tests

koln67:
Unit testing for expired key refactor, added check for empty string in CheckExpiredKeys method (is this necessary?)

### DIFF
--- a/utils/main.go
+++ b/utils/main.go
@@ -43,7 +43,7 @@ type sshExpiration struct {
 // CheckExpiredKey validates whether a key has expired. Keys with invalid expiration formats will result in an
 // error.
 func CheckExpiredKey(key string) error {
-  trimmedKey := strings.Trim(key, " ")
+	trimmedKey := strings.Trim(key, " ")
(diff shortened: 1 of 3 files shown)


utils/main.go
```diff
@@ -43,7 +43,11 @@ type sshExpiration struct {
 // CheckExpiredKey validates whether a key has expired. Keys with invalid expiration formats will result in an
 // error.
 func CheckExpiredKey(key string) error {
-	fields := strings.SplitN(key, " ", 4)
+  trimmedKey := strings.Trim(key, " ")
```
hopkiw, line 46: i think the len(fields) check will catch this condition

koln67: To clarity, would the empty string "    " be considered expired, or not expired?

When len(fields) checks the empty string, it returns a nil error, which seems to mean "not expired."

hopkiw: great point, disregard my comment :)

hopkiw (approved):

hopkiw (approved):

PR #182, opened 2022-09-27
Workload certificate refresh

hopkiw:
* Add service and timer units to run this automatically
* Standardize naming
* Add error handling and file cleanup to refresh binary


jjerger (approved):

gce_workload_cert_refresh/main.go
```diff
@@ -72,8 +72,12 @@ func getMetadata(key string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer res.Body.Close()
 
+	if res.StatusCode == 404 {
+		return nil, fmt.Errorf("HTTP 404")
+	}
+
+	defer res.Body.Close()
```
dorileo, line 80: you might be leaking this Body, move the defer closer to where it got allocated, inside the loop right after you test Do()'s returned error.

gce_workload_cert_refresh/main.go
```diff
@@ -229,19 +235,21 @@ func refreshCreds() error {
 	if err != nil {
 		return fmt.Errorf("Error getting project ID: %v", err)
 	}
-	domain := fmt.Sprintf("%s.svc.id.goog", project)
-	logger.Infof("Rotating workload credentials for domain %s", domain)
 
 	wisMd, err := getMetadata("instance/workload-identities")
 	if err != nil {
-		return fmt.Errorf("Error getting workload-identities: %v", err)
+		logger.Infof("No workload identities found: %v", err)
```
dorileo, line 241: Any special reason to be downgrading this from `Error` to `Info`?

gce_workload_cert_refresh/main.go
```diff
@@ -279,11 +286,24 @@ func refreshCreds() error {
 		return fmt.Errorf("Error creating temporary link: %v", err)
 	}
 
+	oldTarget, err := os.Readlink(symlink)
+	if err != nil {
+		logger.Infof("Error reading existing symlink: %v\n", err)
+		oldTarget = ""
```
dorileo, line 292: No need to initialize `oldTarget` here, `Readlink()` will return an empty string in case of error.

dorileo (commented):
Just minor comments, but the timer & system integration look solid.

gce_workload_cert_refresh/main.go
hopkiw: yep, because this represents the case where workload identity is not present. not being present is a valid scenario, not an error

gce_workload_cert_refresh/main.go
hopkiw: ah i am probably leaning on old guidelines here, but the principle is that functions which return error shouldn't be trusted with any of their other return values

gce_workload_cert_refresh/main.go
hopkiw: good catch - i will update this in future PR. it is low risk right now as this program does not stay resident, but exits on completion.

PR #186, opened 2022-10-20
Updates to gce-workload-cert-refresh

tavishvaidya:
This PR makes the following changes:

* Updates the gce-workload-cert-refresh.timer to run every 10mins.
* Explicitly handles HTTP 412 error while fetching data from MDS. This happens when no valid config was ever provided by the user.
* Handles the edge case where the config provided the very first time are wrong and needs to be communicated to VM via config_status.
* Adds logic to update the config_status in the existing symlink dir to surface latest config errors without rotating the symlink.
* Updates error messages.

### DIFF
--- a/gce_workload_cert_refresh/main.go
+++ b/gce_workload_cert_refresh/main.go
@@ -77,6 +77,10 @@ func getMetadata(key string) ([]byte, error) {
 		return nil, fmt.Errorf("HTTP 404")
 	}
 
+	if res.StatusCode == 412 {
+		return nil, fmt.Errorf("HTTP 412")
+	}
+
 	defer res.Body.Close()
 	md, err := ioutil.ReadAll(res.Body)
 	if err != nil {
@@ -266,7 +270,7 @@ func refreshCreds() error {
 	// that the symlink directory exists and contains the config_status to surface config errors to the VM.
 	defer func() error {
 		if _, err := os.Stat(symlink); os.IsNotExist(err) {
-			logger.Infof("Creating symlink %s", symlink)
+			logger.Infof("Creating new symlink %s", symlink)
 
 			if err := os.Symlink(contentDir, symlink); err != nil {
 				return fmt.Errorf("Error creating symlink link: %v", err)
@@ -288,7 +292,7 @@ func refreshCreds() error {
 
 	wis := WorkloadIdentities{}
 	if err := json.Unmarshal(wisMd, &wis); err != nil {
-		return fmt.Errorf("Error unmarshaling workload trusted root certs: %v", err)
+		return fmt.Errorf("Error unmarshaling workload identities response: %v", err)
 	}
 
 	wtrcs := WorkloadTrustedRootCerts{}
--- a/gce_workload_cert_refresh/main.go
+++ b/gce_workload_cert_refresh/main.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	contentDirPrefix  = "/run/secrets/.workload-spiffe-contents"
+	contentDirPrefix  = "/run/secrets/workload-spiffe-contents"
 	tempSymlinkPrefix = "/run/secrets/workload-spiffe-symlink"
 	symlink           = "/run/secrets/workload-spiffe-credentials"
 )
@@ -276,10 +276,6 @@ func refreshCreds() error {
 		if err := os.Symlink(contentDir, symlink); err != nil {
 			return fmt.Errorf("Error creating symlink: %v", err)
 		}
-		// Write config_status to the newly created symlink immediately.
-		if err := os.WriteFile(fmt.Sprintf("%s/config_status", contentDir), certConfigStatus, 0644); err != nil {
-			return fmt.Errorf("Error writing config_status to contents dir: %v", err)
-		}	
 	}
 
 	// Now get the rest of the content.
--- a/gce_workload_cert_refresh/main.go
+++ b/gce_workload_cert_refresh/main.go
@@ -275,7 +275,6 @@ func refreshCreds() error {
 		return nil
 	}()
 
-
 	// Now get the rest of the content.
 	wisMd, err := getMetadata("instance/workload-identities")
 	if err != nil {
--- a/gce_workload_cert_refresh/main.go
+++ b/gce_workload_cert_refresh/main.go
@@ -331,9 +331,10 @@ func refreshCreds() error {
 	// Clean up previous contents dir.
 	newTarget, err := os.Readlink(symlink)
 	if err != nil {
-		logger.Infof("Error reading new symlink: %v. Unable to remove old symlink target\n", err)
-	} else if oldTarget != "" && oldTarget != newTarget {
-		logger.Infof("Remove old content dir %s", oldTarget)
+		return fmt.Errorf("Error reading new symlink: %v. Unable to remove old symlink target\n", err)
+	}
+	if oldTarget != newTarget {
+		logger.Infof("Removing old content dir %s", oldTarget)
 		if err := os.RemoveAll(oldTarget); err != nil {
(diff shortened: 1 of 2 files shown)


gce_workload_cert_refresh/main.go
```diff
@@ -262,6 +266,25 @@ func refreshCreds() error {
 		return fmt.Errorf("Error writing config_status: %v", err)
 	}
 
+	// Handles the edge case where the config values provided at VM creation may be invalid. This ensures
+	// that the symlink directory exists and contains the config_status to surface config errors to the VM.
+	defer func() error {
```
hopkiw, line 271: a deferred function can't return values; i think you'll need to refactor this a bit to catch the errors you want.

tavishvaidya: Fixed, thanks!

gce_workload_cert_refresh/main.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	contentDirPrefix  = "/run/secrets/workload-spiffe-contents"
+	contentDirPrefix  = "/run/secrets/.workload-spiffe-contents"
```
hopkiw, line 32: no need to do this, this is a convenience for directories and files in the user's home directory. it would be confusing to use that convention here.

gce_workload_cert_refresh/main.go
```diff
@@ -77,6 +77,10 @@ func getMetadata(key string) ([]byte, error) {
 		return nil, fmt.Errorf("HTTP 404")
 	}
 
+	if res.StatusCode == 412 {
```
hopkiw, line 80: since this is an abuse of this code, let's add a comment about why we're treating it specially and when to expect it.

gce_workload_cert_refresh/main.go
```diff
@@ -223,9 +227,8 @@ func main() {
 	defer logger.Infof("Done")
 
 	// TODO: prune old dirs
-
 	if err := refreshCreds(); err != nil {
-		logger.Fatalf(err.Error())
+		logger.Errorf("Error refreshCreds: %v", err.Error())
```
hopkiw, line 231: what is the reason to change from fatal to error? fatal will cause the process to exit with an error code, which will be reflected by the systemd service status, which is why it was chosen.

gce_workload_cert_refresh/main.go
```diff
@@ -244,6 +247,16 @@ func refreshCreds() error {
 		return nil
 	}
 
+	// Write the config status immediately to the current symlink if it exists. This allows for communicating 
+	// new config errors. E.g., a user can provide wrong config values multiple times. This requires update to 
+	// the config_status without rotating the symlink because it may contain valid credentials.
```
hopkiw, line 252: this comment is unclear

gce_workload_cert_refresh/main.go
```diff
@@ -244,6 +247,16 @@ func refreshCreds() error {
...
+	if _, err := os.Stat(symlink); err == nil {
+		logger.Infof("Writing config_status to existing symlink")
+		if err := os.WriteFile(fmt.Sprintf("%s/config_status", symlink), certConfigStatus, 0644); err != nil {
```
hopkiw, line 255: what is the intent of writing here, then writing again later? the only steps that could fail in between the two writes are creating the new per-run contents dir and then writing the config_status there again. suggest deleting this entire block.

gce_workload_cert_refresh/main.go
```diff
@@ -262,6 +275,20 @@ func refreshCreds() error {
 		return fmt.Errorf("Error writing config_status: %v", err)
 	}
 
+	// Handles the edge case where the config values provided at VM creation may be invalid. This ensures
```
hopkiw, line 278: i think at first run is clearer than at VM creation, because this can be enabled post-boot, right?

gce_workload_cert_refresh/main.go
```diff
@@ -262,6 +275,20 @@ func refreshCreds() error {
...
+	// that the symlink directory exists and contains the config_status to surface config errors to the VM.	
+	if _, err := os.Stat(symlink); os.IsNotExist(err) {
+		logger.Infof("Creating new symlink %s", symlink)
+
+		if err := os.Symlink(contentDir, symlink); err != nil {
+			return fmt.Errorf("Error creating symlink link: %v", err)
+		}
+		// Write config_status to the newly created symlink immediately.
+		if err := os.WriteFile(fmt.Sprintf("%s/config_status", symlink), certConfigStatus, 0644); err != nil {
```
hopkiw, line 287: it's not clear why we have now a *third* write of config_status. we should also never write through the symlink. in this codebase, we always reference the per-run directory, and we do symlink operations separately.

gce_workload_cert_refresh/main.go
```diff
@@ -262,6 +275,20 @@ func refreshCreds() error {
 		return fmt.Errorf("Error writing config_status: %v", err)
 	}
 
+	// Handles the edge case where the config values provided at VM creation may be invalid. This ensures
+	// that the symlink directory exists and contains the config_status to surface config errors to the VM.	
+	if _, err := os.Stat(symlink); os.IsNotExist(err) {
+		logger.Infof("Creating new symlink %s", symlink)
+
+		if err := os.Symlink(contentDir, symlink); err != nil {
+			return fmt.Errorf("Error creating symlink link: %v", err)
```
hopkiw, line 284: change 'symlink link' to 'symlink'

gce_workload_cert_refresh/main.go
```diff
@@ -312,11 +339,16 @@ func refreshCreds() error {
...
 	}
 
-	if oldTarget != "" {
-		logger.Infof("Remove old content dir %s", oldTarget)
-		if err := os.RemoveAll(oldTarget); err != nil {
-			return fmt.Errorf("Failed to remove old symlink target: %v", err)
+	newTarget, err := os.Readlink(symlink)
+	if err == nil {
+		if oldTarget != "" && oldTarget != newTarget {
+			logger.Infof("Remove old content dir %s", oldTarget)
```
hopkiw, line 345: several different ways we could approach this to make it readable, but i think the primary issue is this kind of code is a product of the strategy used above. suggest reviewing feedback above, revisiting strategy, then this section may become more obvious.

hopkiw (changes requested):

gce_workload_cert_refresh/main.go
hopkiw: yes, please revert

gce_workload_cert_refresh/main.go
```diff
@@ -262,6 +268,20 @@ func refreshCreds() error {
...
+	// Handles the edge case where the config values provided for the first time may be invalid. This ensures
+	// that the symlink directory alwasys exists and contains the config_status to surface config errors to the VM.
+	if _, err := os.Stat(symlink); os.IsNotExist(err) {
+		logger.Infof("Creating new symlink %s", symlink)
+
+		if err := os.Symlink(contentDir, symlink); err != nil {
+			return fmt.Errorf("Error creating symlink: %v", err)
+		}
+		// Write config_status to the newly created symlink immediately.
+		if err := os.WriteFile(fmt.Sprintf("%s/config_status", contentDir), certConfigStatus, 0644); err != nil {
```
hopkiw, line 280: this is already done above on line 267

gce_workload_cert_refresh/main.go
hopkiw: this was logic originally to handle the first creation case, which you've changed in this PR. oldtarget can now never be "", since you unconditionally create the symlink on line 276. what are you trying to do here, maybe i can suggest an amendment?

hopkiw (changes requested):

hopkiw (approved):

PR #187, opened 2022-10-27
Implement validation in accounts_unix.go

ElijahPepe:
The createUser() and addUserToGroup() functions don't validate what the value of username, user, or group is, leading to strings.Fields() executing any flags present in either of those strings as flags. Group and user names can't contain the space character on POSIX-style systems, so validating the parameters of those functions by disallowing whitespace characters (implemented here through RegEx) presents no complications.

### DIFF
--- a/google_guest_agent/accounts_unix.go
+++ b/google_guest_agent/accounts_unix.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


hopkiw:
Thank you @ElijahPepe for the contribution. Coincidentally, we have a pending change for the same issue. I am closing this PR in favor of that.

PR #188, opened 2022-10-28
Validate user key for whitespace chars

koln67:
Add validation and testing for whitespace in user string. If whitespace is detected in the user string, it should be considered an invalid user string because it could be exploiting a security loophole.

### DIFF
--- a/utils/main.go
+++ b/utils/main.go
@@ -93,9 +93,9 @@ func CheckExpired(expireOn string) (bool, error) {
 // allowed in a username string, returns an error if any such characters are
 // detected, nil otherwise.
 // Currently, the only banned characters are whitespace characters.
-func ValidateUserKey(user string) (error) {
+func ValidateUserKey(user string) error {
 	whiteSpaceRegexp, _ := regexp.Compile("\\s")
-	
+
 	hasWhiteSpace := whiteSpaceRegexp.MatchString(user)
 	if hasWhiteSpace {
 		return errors.New("Invalid username - whitespace detected")
--- a/utils/main.go
+++ b/utils/main.go
@@ -19,6 +19,7 @@ package utils
 import (
 	"encoding/json"
 	"errors"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


utils/main.go
```diff
@@ -88,6 +89,20 @@ func CheckExpired(expireOn string) (bool, error) {
 
 }
 
+// ValidateUserKey checks for the presence of a characters which should not be
+// allowed in a username string, returns an error if any such characters are
+// detected, nil otherwise.
+// Currently, the only banned characters are whitespace characters.
+func ValidateUserKey(user string) (error) {
+	whiteSpaceRegexp, _ := regexp.Compile("\\s")
```
dorileo, line 97: Do you need to use `regexp` for this purpose? Wont `strings.Contains()` be more efficient?

utils/main_test.go
```diff
@@ -83,3 +83,25 @@ func TestCheckExpiredKey(t *testing.T) {
...
+		{"user -g 27", false},
+		{"user\t-g", false},
+		{"user\n-g", false},
+		{"username\t-g\n27", false},
+	}
+	for _, tt := range table {
+		err := ValidateUserKey(tt.user)
+		isValid := err != nil
+		if isValid != tt.valid {
+			t.Errorf("ValidateUserKey(%s) incorrect return: valid: %t - want valid: %t", tt.user, isValid, tt.valid)
```
dorileo, line 104: I'd rephrase it as: `"Invalid ValidateUserKey(%s) return: expected: %t - got: %t"`

dorileo (commented):
Minor comments...

utils/main.go
koln67: Does strings.Contains() have a way to check for any whitespace character? 

Strictly speaking, it seems like anything which falls under unicode.IsSpace() should not be in the user string.

dorileo: I see, looks good then.

utils/main.go
```diff
@@ -88,6 +89,20 @@ func CheckExpired(expireOn string) (bool, error) {
...
 
+// ValidateUserKey checks for the presence of a characters which should not be
+// allowed in a username string, returns an error if any such characters are
+// detected, nil otherwise.
+// Currently, the only banned characters are whitespace characters.
+func ValidateUserKey(user string) error {
+	whiteSpaceRegexp, _ := regexp.Compile("\\s")
+
+	hasWhiteSpace := whiteSpaceRegexp.MatchString(user)
+	if hasWhiteSpace {
```
hopkiw, line 100: these lines can be combined

utils/main_test.go
koln67: Changed

utils/main.go
koln67: Changed: now reads with hasWhiteSpace == true for readability

utils/main.go
```diff
@@ -88,6 +89,19 @@ func CheckExpired(expireOn string) (bool, error) {
...
 }
 
+// ValidateUserKey checks for the presence of a characters which should not be
+// allowed in a username string, returns an error if any such characters are
+// detected, nil otherwise.
+// Currently, the only banned characters are whitespace characters.
+func ValidateUserKey(user string) error {
+	whiteSpaceRegexp, _ := regexp.Compile("\\s")
+
+	if hasWhiteSpace := whiteSpaceRegexp.MatchString(user); hasWhiteSpace == true {
```
dorileo, line 99: Are you using `hasWhiteSpace` elsewhere?

Why not simply doing the following(I guess this is what @hopkiw meant):

```go
if whiteSpaceRegexp.MatchString(user) {
```

koln67: fixed

hopkiw (approved):

PR #192, opened 2022-12-15
Auth key command

koln67:
When a user ssh key is changed on a VM, previously the changes were being written to an authorized keys file. 

After this change, an AuthorizedKeysCommand binary will be used instead of the authorized keys file.


google_guest_agent/non_windows_accounts.go
```diff
@@ -320,6 +321,40 @@ func readGoogleUsersFile() (map[string]string, error) {
...
+	proposed := filterAuthorizedKeyLines(string(sshConfig))
+	if proposed == string(sshConfig) {
+		return nil
+	}
+	file, err := os.OpenFile("/etc/ssh/sshd_config", os.O_WRONLY|os.O_TRUNC, 0777)
+	if err != nil {
+		return err
+	}
+	defer closeFile(file)
+	file.WriteString(proposed)
```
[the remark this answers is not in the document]
dorileo: Are you rewriting `/etc/ssh/sshd_config` with only the 2 lines returned by `filterAuthorizedKeyLines()`?

google_guest_agent/non_windows_accounts.go
```diff
@@ -320,6 +321,40 @@ func readGoogleUsersFile() (map[string]string, error) {
...
+	return nil
+}
+
+// Helper function to set AuthorizedKeysCommand, should only run once in entire program execution
+func filterAuthorizedKeyLines(contents string) string {
```
dorileo, line 344: Don't you want to search for key tokens only (i.e AuthorizedKeysCommand and AuthorizedKeysCommandUser) and then return the actual matching lines?

google_guest_agent/non_windows_accounts.go
```diff
@@ -320,6 +321,40 @@ func readGoogleUsersFile() (map[string]string, error) {
 	return res, nil
 }
 
+// Enables AuthorizedKeysCommand to be used instead of an AuthorizedKeysFile
```
dorileo, line 324: I don't think this is actually doing what this comment says. It's reading the `sshd_config` file trying to find a token `AuthorizedKeysCommand` and then rewriting the file with  	`#AuthorizedKeysCommand /usr/bin/google_authorized_keys_guest` and `#AuthorizedKeysCommandUser root` meaning you are disabling these two configurations - but erasing all the remaining configs.

dorileo (changes requested):

google_guest_agent/non_windows_accounts.go
koln67: I think it's checking each existing line, and adds any line back to the file other than the existing "AuthorizedKeysCommand" and "AuthorizedKeysCommandUser" line. 

If the check was 
"if strings.Contains(line, "AuthorizedKeysCommand")" then I think it would delete all lines except those two lines. Right now the check is 
"if !strings.Contains(line, "AuthorizedKeysCommand")."

dorileo: Right, I misread the conditional statement. Thanks for the correction.

google_guest_agent/non_windows_accounts.go
```diff
@@ -320,6 +321,40 @@ func readGoogleUsersFile() (map[string]string, error) {
...
+func writeSSHDConfigFile() error {
+	sshConfig, err := ioutil.ReadFile("/etc/ssh/sshd_config")
+	if err != nil {
+		return err
+	}
+	proposed := filterAuthorizedKeyLines(string(sshConfig))
+	if proposed == string(sshConfig) {
+		return nil
+	}
+	file, err := os.OpenFile("/etc/ssh/sshd_config", os.O_WRONLY|os.O_TRUNC, 0777)
```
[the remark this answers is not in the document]
koln67: I'm following the convention in this line: 

https://github.com/koln67/guest-agent/blob/7905b623a6911e7ad35dbfac2242b7c0cbfac02a/google_guest_agent/oslogin.go#L178

Much of the file writing code is similar to what is in that file, so there might be some repetitive code.

google_guest_agent/non_windows_accounts.go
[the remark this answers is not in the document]
koln67: checked the return value

google_guest_agent/non_windows_accounts.go
```diff
@@ -111,6 +111,11 @@ func (a *accountsMgr) set() error {
 	if sshKeys == nil {
 		logger.Debugf("initialize sshKeys map")
 		sshKeys = make(map[string][]string)
+		// only need to write to SSHD once in program execution
+		if err := writeSSHDConfigFile(); err != nil {
+			logger.Errorf("Error updating SSH config: %v.", err)
+		}
+		systemctlStart("sshd")
```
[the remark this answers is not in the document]
koln67: checked

google_guest_agent/non_windows_accounts.go
```diff
@@ -320,6 +321,40 @@ func readGoogleUsersFile() (map[string]string, error) {
 	return res, nil
 }
 
+// Enables AuthorizedKeysCommand to be used instead of an AuthorizedKeysFile
+func writeSSHDConfigFile() error {
+	sshConfig, err := ioutil.ReadFile("/etc/ssh/sshd_config")
```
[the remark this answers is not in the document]
koln67: replaced

packaging/google-guest-agent.spec
```diff
@@ -38,7 +38,7 @@ Contains the Google guest agent binary.
 %autosetup
 
 %build
-for bin in google_guest_agent google_metadata_script_runner gce_workload_cert_refresh; do
+for bin in google_authorized_keys_guest google_guest_agent google_metadata_script_runner gce_workload_cert_refresh; do
```
hopkiw, line 41: ok, we can't rename this directory, or the windows builds will break. i think you should restore the name but add rename steps to the debian rules and rpm spec. or else we would want to coordinate with the windows releases to ensure they don't break with the rename.

google_guest_agent/non_windows_accounts.go
```diff
@@ -320,6 +323,43 @@ func readGoogleUsersFile() (map[string]string, error) {
...
+	defer closeFile(file)
+	_, err = file.WriteString(proposed)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Helper function to set AuthorizedKeysCommand, should only run once in entire program execution
+func filterAuthorizedKeyLines(contents string) string {
```
hopkiw, line 349: rather than cloning and editing the file editing functions, can we refactor the existing function so it can be used for both oslogin and metadata enablement modes?

google_guest_agent/non_windows_accounts.go
```diff
@@ -111,6 +111,13 @@ func (a *accountsMgr) set() error {
 	if sshKeys == nil {
 		logger.Debugf("initialize sshKeys map")
 		sshKeys = make(map[string][]string)
+		// only need to write to SSHD once in program execution
+		if err := writeSSHDConfigFile(); err != nil {
+			logger.Errorf("Error updating SSH config: %v.", err)
+		}
+		if err := systemctlStart("sshd"); err != nil {
```
hopkiw, line 118: it would be `restart`, not `start`. but don't we already restart sshd?

packaging/google-guest-agent.spec
hopkiw: oh, and this PR is currently missing the debian updates

hopkiw (changes requested):

packaging/google-guest-agent.spec
koln67: By rename steps, would this be sufficient to do a similar renaming in this file (without changing the directory name)? 

install -p -m 0755 google_guest_agent/google_authorized_keys %{buildroot}%{_bindir}/google_authorized_keys_guest

hopkiw: i think it should work; but you should be able to test it out yourself.

google_guest_agent/non_windows_accounts.go
koln67: writeSSHDConfigFile() can probably be removed, and we can add an argument to the existing writeSSHConfigFile() function for this purpose.

I do not see any existing function which does what filterAuthorizedKeyLines() does, however. Is the suggestion to merge this functionality with something such as updateSSHConfig()? 

I'm thinking about calling something like filterAuthorizedKeyLines() instead of updateSSHConfig() in the metadata mode.

google_guest_agent/non_windows_accounts.go
koln67: this function finds the AuthorizedKeysCommand and AuthorizedKeysCommandUser lines, and replaces them to use the google_authorized_keys_guest binary

All other lines should remain unchanged.

google_guest_agent/non_windows_accounts.go
koln67: removed

packaging/google-guest-agent.spec
koln67: tested and seems to work

dorileo:
this PRs is pretty outdated and we have plans for consolidating these tools in a broader sense. Closing it for now.

PR #193, opened 2022-12-22
use stricter regex for username validation

koln67:
Due to other security threats, such as the possibility of path traversal attacks, this is a proposed change to ensure robustness of usernames from a variety of possible security threats.

### DIFF
--- a/utils/main.go
+++ b/utils/main.go
@@ -94,10 +94,14 @@ func CheckExpired(expireOn string) (bool, error) {
 // detected, nil otherwise.
 // Currently, the only banned characters are whitespace characters.
 func ValidateUserKey(user string) error {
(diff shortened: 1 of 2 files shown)


utils/main.go
```diff
@@ -94,10 +94,14 @@ func CheckExpired(expireOn string) (bool, error) {
 // detected, nil otherwise.
 // Currently, the only banned characters are whitespace characters.
 func ValidateUserKey(user string) error {
-	whiteSpaceRegexp, _ := regexp.Compile("\\s")
+	if len(user) > 32 {
+		return errors.New("Invalid username - exceeds maximum length")
+	}
+	// recommended linux username regex according to the manual
```
hopkiw, line 100: what manual do you mean?

PR #198, opened 2023-01-27
Allow a comment part of a pub ssh key to have an arbitrary format

vorakl:
For a public ssh key in the "OpenSSH format", simply represented as
SIGALGS KEY COMMENT
this patch allows a COMMENT part of a key to be in a free form while keeping an expiration declaration in a valid state.

Basically, the validation function checks if the 1st part of a comment equals to "google-ssh" then there should be a 2nd part and it also should be a valid JSON document with an expected content.
But, if the 1st part of a comment is not "google-ssh", it passes the validation and allows a key to have a comment part of a key with multiple spaces and have a free format.


dorileo:
hi @vorakl, please consider the following. Split `CheckExpiredKey()` into a second function (maybe `ParseSSHKey()` ?) dedicated only to *parse* the ssh key, breaking it into user, user's key file content, and google-ssh object. Then add a unit tests to `ParseSSHKey()`.

google_authorized_keys/main.go
```diff
@@ -99,6 +99,10 @@ func parseSSHKeys(username string, keys []string) []string {
 		}
 
 		user, keyVal, err := utils.GetUserKey(key)
+		if err == nil {
```
dorileo, line 102: Why this check here? can't you simply move `utils.ValidateUserKey(user, keyVal)` after the `if err != nil {` block?

google_guest_agent/non_windows_accounts.go
```diff
@@ -204,6 +204,10 @@ func getUserKeys(mdkeys []string) map[string][]string {
 		trimmedKey := strings.Trim(mdkeys[i], " ")
 		if trimmedKey != "" {
 			user, keyVal, err := utils.GetUserKey(trimmedKey)
+			if err == nil {
```
dorileo, line 207: Same here.

utils/main.go
```diff
@@ -102,29 +110,40 @@ func ValidateUserKey(user string) error {
...
 }
 
-// GetUserKey takes a string and determines if it is a valid SSH key and returns
-// the user and key if valid, nil otherwise.
+// GetUserKey returns a user and a SSH key if a rawKey has a correct format, nil otherwise.
+// It doesn't validate entries.
 func GetUserKey(rawKey string) (string, string, error) {
-	key := strings.Trim(rawKey, " ")
-	if key == "" {
+	tmpKey := strings.Trim(rawKey, " ")
```
dorileo, line 116: Why do you need tmp key here? We're chaining the string transformation.

dorileo (changes requested):

google_authorized_keys/main.go
vorakl: Nope, I can't. The GetUserKey function used to be a combination of getting and validating parts. But, after splitting them into two functions, we need not only call them separately, but also evaluate errors separately either. Errors can be raised independently here. I'd expect you ask me why there is no `else` ;) So, else cannot be there either.

The idea here is that GetUserKey can fail by its own reasons, and we need to check this condition before calling a validation function, which in turn, won't make any sense if there was an error.  In this case, a 2nd if-statement will lead to get a true state and execute continue (in other words, it will ignore that key). 

Although, if GetUserKey was successful, and only in this case, we can validate a user and a key. Again, if validation fails, it continue (ignore) the key.

The short answer, if I run a validation on a previously error condition, it would not only mask the error (because validation could succeed), but also doesn't have any sense in general (because the getter has already failed at that moment)

utils/main.go
vorakl: You're right, I'll rewrite it.

vorakl: Done.

dorileo (approved):

PR #206, opened 2023-04-26
Add a support of TrustedUserCAKeys into sshd configuration

vorakl:


google_guest_agent/oslogin.go
```diff
@@ -198,6 +198,7 @@ func updateSSHConfig(sshConfig string, enable, twofactor, skey bool) string {
 		}
 	}
 	authorizedKeysUser := "AuthorizedKeysCommandUser root"
+	trustedUserCAKeys := "TrustedUserCAKeys /etc/ssh/trustedca.pub"
```
dorileo, line 201: What happens if the system doesn't have such a file: `/etc/ssh/trustedca.pub` ?

vorakl: I tested this. It works perfectly fine. Even more, when you add the file, SSHD finds it and starts using with no restart needed

vorakl: As I explained this on a sync meeting, the change is totally independent  to the rest of modifications and is safe to have it merged already.

dorileo:
@vorakl why does it need to be a global configuration?

vorakl:
@dorileo, We need this configuration in the global file because of a number of reasons:
- OS Login users are not local
- OS Login users are not known in advance
- home directories are created by PAM module as the last action, but this happens after SSHD checked ssh certificates

Local configuration in the user's home dirs are not possible to use in the existing OS Login design

chandrasekhar1996:
@vorakl @dorileo What is the motivation behind adding this line in sshd_config? Is it the plan that gues-agent will write this line in the future?
If I already have an entry for TrustedUserCAKeys and I enable os-login then it breaks my setup since os-login adds these lines to the beginning of the files.

Issue #209, opened 2023-05-09
Vendored unix module fails to build due to Go version too old in go.mod


glaubitz:
I was just trying to update the `guest-agent` package in openSUSE and the build fails with:

```
[   18s] + go build '-ldflags=-s -w -X main.version=20230426.00' -mod=vendor
[   18s] # golang.org/x/sys/unix
[   18s] ../vendor/golang.org/x/sys/unix/syscall.go:83:16: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
[   18s] ../vendor/golang.org/x/sys/unix/syscall_linux.go:2271:9: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
[   18s] ../vendor/golang.org/x/sys/unix/syscall_unix.go:118:7: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
[   18s] ../vendor/golang.org/x/sys/unix/sysvshm_unix.go:33:7: unsafe.Slice requires go1.17 or later (-lang was set to go1.16; check go.mod)
[   23s] error: Bad exit status from /var/tmp/rpm-tmp.Pw9TCI (%build)
```

due to the Go version in `go.mod` being too old.

Please bump the Go version in `go.mod` to 1.17 or later.

vorakl:
Thank you for the filing this issue!
The Go version has been updated to 1.17

glaubitz:
Thanks! Could you tag a new release so I can use this version for packaging?

vorakl:
Sure, thanks for pointing this out!
It used to add release tags automatically for every new commit on the main branch.
I'll take a look why it didn't happen this time.

vorakl:
A new release tag is available: 	20230510.00

vorakl:
I can't help with that, there are other guys. I'll ping them.

PR #210, opened 2023-05-09
Update Go version

vorakl:
https://github.com/GoogleCloudPlatform/guest-agent/issues/209


dorileo:
please, let us comply with the minimum required (1.17) and reduce the surface of potential side effects.

vorakl:
Sounds good to me

PR #217, opened 2023-07-06
snapshot: fix request handling error

dorileo:
In the snapshot request handling code we get a duplicate pre or post request (a request to previously seen snapshot) we ignore it and don't produce a response - the code is assuming always a valid response as of now, causing a vm resource's exhaustion (both memory & cpu).

With a nil response the rpc client will fail to marshal the response and return error. We don't want to call HandleResponsesFromGuest() with an invalid/nil response in the first place (this is a known condition to error) instead we want to ignore this condition.

Additionally this patch fixes other aspects of the handling code, we moved the cancel() callback to be executed in error or success conditions and established a max attempt limit - in that case if any other error condition is met we don't exhaust the system trying to process it even if it's not a platform issue.


dorileo:
I'm keeping this one on hold until consensus is met.

PR #218, opened 2023-07-10
metadata: moving to its own package

dorileo:
Creating a separation between the core implementation and metadata will give us the flexibility to move internals as required without breaking the agent and managers specifics.

No behavior has changed, the current coverage shoud be enough for now.

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -12,7 +12,7 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-package main
+package metadata
 
 import (
 	"bytes"
@@ -25,7 +25,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/GoogleCloudPlatform/guest-agent/utils"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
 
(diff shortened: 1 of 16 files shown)


metadata/metadata.go
```diff
@@ -286,6 +261,21 @@ func getMetadata(ctx context.Context, hang bool) (*metadata, error) {
...
+	var ret Descriptor
 	return &ret, json.Unmarshal(md, &ret)
 }
+
+// WriteGuestAttributes does a put call to mds changing a guest attribute value.
+func WriteGuestAttributes(key, value string) error {
+	logger.Debugf("write guest attribute %q", key)
+	client := &http.Client{Timeout: defaultTimeout}
+	finalURL := metadataURL + "instance/guest-attributes/" + key
+	req, err := http.NewRequest("PUT", finalURL, strings.NewReader(value))
```
ChaitanyaKulkarni28, line 273: this would create new context on each new http request, why not create it top of the call chain and accept it here as an argument?

metadata/metadata.go
```diff
@@ -244,12 +217,14 @@ func updateEtag(resp *http.Response) bool {
 	return etag != oldEtag
 }
 
-func watchMetadata(ctx context.Context) (*metadata, error) {
-	return getMetadata(ctx, true)
+// Watch runs a longpool on metadata server.
```
ChaitanyaKulkarni28, line 220: nit: `long poll` here and all other usages?

metadata/metadata.go
```diff
@@ -244,12 +217,14 @@ func updateEtag(resp *http.Response) bool {
...
+func Watch(ctx context.Context) (*Descriptor, error) {
+	return Get(ctx, true)
 }
 
-func getMetadata(ctx context.Context, hang bool) (*metadata, error) {
-	logger.Debugf("getMetadata, %t", hang)
+// Get does a metadata call, if hang is set to true then it will do a longpool.
+func Get(ctx context.Context, hang bool) (*Descriptor, error) {
+	logger.Debugf("Get, %t", hang)
```
ChaitanyaKulkarni28, line 227: this will log just `Get, true/false` can we log something like `Invoking Get metadata, wait for change: true`

metadata/metadata.go
dorileo: This function was just moved from one place to another without an actual rewrite or improvement. It will certainly have a rewrite anytime soon.

metadata/metadata.go
dorileo: Harmless, I'll have it done for this PR. Thank you.

metadata/metadata.go
dorileo: done.

PR #219, opened 2023-07-11
Add (disabled) telemetry logic to GuestAgent

adjackura:
Recording telemetry is disabled by default for now. It will only be enabled if `disable-guest-telemetry` is set to false in metadata.


google_guest_agent/osinfo/osinfo_unix.go
```diff
@@ -12,42 +12,22 @@
...
-	}
-	return ret
-}
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	"golang.org/x/sys/unix"
+)
 
-func parseOSRelease(osRelease string) (release, error) {
-	var ret release
+func parseOSRelease(osRelease string) OSInfo {
```
dorileo, line 29: These functions here where you are handling errors internally (meaning not returning error anymore), I'd like to have them reporting the error and allowing the caller decide what to do with it, even if it's just logging and not aborting, as a "specialized code" it shouldn't know about the "app"/guest-agent flow.

dorileo: meaning, propagating the error up to the "core" guest-agent code.

google_guest_agent/osinfo/osinfo_unix.go
```diff
@@ -130,16 +114,27 @@ func parseID(id string) string {
...
-	if runtime.GOOS == "linux" {
-		releaseFile, err := ioutil.ReadFile("/etc/os-release")
-		if err == nil {
-			return parseOSRelease(string(releaseFile))
-		}
+// Get returns OSInfo on the running system.
+func Get() OSInfo {
+	var osInfo OSInfo
+
+	releaseFile, err := ioutil.ReadFile("/etc/os-release")
```
dorileo, line 121: I'd break this block into its own util function as well, receiving a list of possible release file paths.

google_guest_agent/osinfo/osinfo_unix.go
```diff
@@ -12,42 +12,22 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-package main
+//go:build unix
```
dorileo, line 15: we don't need the directive/pragma here, let's just use the file naming convention. (I'm migrating it wherever we can leaving the pragma only for cases where the convention will suit).

google_guest_agent/osinfo/osinfo_unix_test.go
```diff
@@ -12,7 +12,9 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
-package main
+//go:build unix
```
dorileo, line 15: ditto.

google_guest_agent/osinfo/osinfo_unix_test.go
```diff
@@ -22,14 +24,14 @@ import (
 func TestParseSystemRelease(t *testing.T) {
 	tests := []struct {
 		file string
-		want release
+		want OSInfo
 	}{
-		{"Red Hat Enterprise Linux Server release 6.10 (Santiago)", release{os: "rhel", version: ver{6, 10, 0, 2}}},
-		{"Red Hat Enterprise Linux Server release 6.10.1", release{os: "rhel", version: ver{6, 10, 1, 3}}},
-		{"CentOS Linux release 7.6.1810 (Core)", release{os: "centos", version: ver{7, 6, 1810, 3}}},
+		{"Red Hat Enterprise Linux Server release 6.10 (Santiago)", OSInfo{OS: "rhel", Version: Ver{6, 10, 0, 2}}},
```
dorileo, line 29: I'd love to see a test exercising error conditions as much as feasible.

google_guest_agent/osinfo/osinfo_windows.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+var (
+	version                     = windows.NewLazySystemDLL("version.dll")
+	procGetFileVersionInfoSizeW = version.NewProc("GetFileVersionInfoSizeW")
+	procGetFileVersionInfoW     = version.NewProc("GetFileVersionInfoW")
+	procVerQueryValueW          = version.NewProc("VerQueryValueW")
+)
+
+// https://msdn.microsoft.com/en-us/library/windows/desktop/ms647464(v=vs.85).aspx
```
dorileo, line 37: Can you please describe the rationale rather than linking to the api doc?

google_guest_agent/osinfo/osinfo_windows.go
```diff
@@ -0,0 +1,175 @@
...
+	"syscall"
+	"unsafe"
+
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
+)
+
+var (
+	version                     = windows.NewLazySystemDLL("version.dll")
```
dorileo, line 31: Can you document these variables? giving a bit more context to what they mean.

google_guest_agent/osinfo/osinfo_windows.go
```diff
@@ -0,0 +1,175 @@
...
+	// https://msdn.microsoft.com/en-us/library/windows/desktop/dd318693(v=vs.85).aspx
+	t[0], t[1] = trans[1], trans[0]
+	// 16-bit code page ID little endian
+	// https://msdn.microsoft.com/en-us/library/windows/desktop/dd317756(v=vs.85).aspx
+	t[2], t[3] = trans[3], trans[2]
+
+	return fmt.Sprintf("%x", t), nil
+}
+
+// https://msdn.microsoft.com/en-us/library/windows/desktop/ms647464(v=vs.85).aspx
```
dorileo, line 67: ditto.

google_guest_agent/osinfo/osinfo_windows.go
```diff
@@ -0,0 +1,175 @@
...
+	langCodePage, err := getTranslation(info)
+	if err != nil {
+		return "", "", fmt.Errorf("getTranslation() error: %v", err)
+	}
+
+	return getVersion(info, langCodePage)
+}
+
+type win32OperatingSystem struct {
+	Caption, Version string
```
dorileo, line 138: Can you also describe what these fields mean in the context of the osinfo implementation for windows?

google_guest_agent/telemetry/telemetry.go
```diff
@@ -0,0 +1,100 @@
...
+	return base64.StdEncoding.EncodeToString(data)
+}
+
+// Record records telemetry data.
+func Record(ctx context.Context, d Data) error {
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+	}
+
+	req, err := http.NewRequest("GET", metadataURL, nil)
```
dorileo, line 89: Can you move this mds call to the metadata package and just wrap it from here? You can reuse the metadata client in the `main.go` passing it down to this function - later on I plan to improve the call pattern of the mds client but for now it should be enough.

google_guest_agent/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestRecordTelemetry(t *testing.T) {
+	var got http.Header
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
dorileo, line 26: Can you start a "fake" implementation in the metadata package in way that you can use it here? You don't need to make it "perfect", just make sure it's centralized with the metadata package so we can start improving/increasing its coverage in a single place.

google_guest_agent/main.go
```diff
@@ -161,6 +159,31 @@ func run(ctx context.Context) {
 
 	agentInit(ctx)
 
+	if newMetadata == nil {
```
dorileo, line 162: I were to suggest the following `"I guess you don't depend on anything done on behalf of `agentInit()` to do the telemetry calls, right? in that case I'd suggest moving it above the `agentInit()` just to make sure we express this info."` but thinking twice we might not want to block initialization for the sake of telemetry, said that I'd suggest just commenting that `// Telemetry is not agent initialization dependent.`. I'll be touching this part of the code any time soon and I want to make sure I can move stuff to the right place when I get to it.

google_guest_agent/osinfo/osinfo_unix.go
adjackura: unix is not an os type (like linux or windows) so file names don't work, build directives are required

google_guest_agent/osinfo/osinfo_windows.go
adjackura: Added the links to these procedure calls in msft docs, that seems pretty self explanatory?

google_guest_agent/telemetry/telemetry.go
adjackura: What part should I wrap? I need to add headers, should I update the existing Get call to add optional headers? I don't actually care about metadata in this function at all (or any of the other code in that function), just need to make some call that includes these headers, updating Get seemed more complicated as that impacted any other callers, also it would include an unneeded json unmarhal.

google_guest_agent/main.go
adjackura: Yeah, added some more comments, let me know if I should provide even more data. This is a bit odd/specific to the telemetry use case

google_guest_agent/osinfo/osinfo_unix.go
adjackura: I broke it out, but the paths are specific to the parsing functions.

google_guest_agent/osinfo/osinfo_unix.go
adjackura: These are private now, very different from before so I can't pass the errors up to core code. I did move the error handling to the Get function though so that we can test the error cases here.

I could make Get a bit more complicated and return concatenated errors (Get won't return immediately on error, but instead get adding to the potential errors) if you would prefer the caller of Get to handle the logic

dorileo:
Did one more pass, the PR is looking solid in my opinion. We just need to get sorted the disagreements around metadata and tests.

google_guest_agent/osinfo/osinfo_windows.go
dorileo: I wanted the code to be self-contained - to not depend on external resources/links. But we can neglect it for now.

PR #220, opened 2023-07-14
metadata: add metadata client interface

dorileo:
Change how the metadata layer/subsystem presents itself and how the guest agent core implementation interacts with it.

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -28,15 +28,35 @@ import (
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
 
-const defaultEtag = "NONE"
+const (
+	defaultMetadataURL = "http://169.254.169.254/computeMetadata/v1/"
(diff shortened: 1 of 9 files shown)


metadata/metadata.go
```diff
@@ -270,16 +286,15 @@ func get(ctx context.Context, hang bool) (*Descriptor, error) {
 }
 
 // WriteGuestAttributes does a put call to mds changing a guest attribute value.
-func WriteGuestAttributes(key, value string) error {
+func (c *Client) WriteGuestAttributes(key, value string) error {
```
ChaitanyaKulkarni28, line 289: since we're already updating this, can we also change its signature to accept context from caller and use it while making HTTP calls?

dorileo: surething. will do. Thanks for catching this one.

dorileo: done.

PR #221, opened 2023-07-18
events: introducing a events handling subsystem

dorileo:
The events subsystem offers an unified interface between guest agent and systems and services events.

The subsystem defines a Watcher interface implemented by such listeners, when a event occurs the watcher reports it back to the events subsystem that will then distribute "events notifications" to registered callbacks or subscribers.

### DIFF
--- a/google_guest_agent/events/events.go
+++ b/google_guest_agent/events/events.go
@@ -1,0 +1,271 @@
+//  Copyright 2023 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package events is a events processing layer.
+package events
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/events/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+var (
+	// availableWatchers mapps all kown available event watchers.
+	availableWatchers = make(map[string]Watcher)
+)
+
+// Watcher defines the interface between the events manager and the actual
+// watcher implementation.
+type Watcher interface {
+	// ID returns the watcher id.
+	ID() string
+	// Run implements the actuall "listening" strategy and emits a event "signal".
+	// It must return:
+	//   - [bool] if the watcher should renew(run again).
+	//   - [string] the event type id.
+	//   - [interface{}] a event context data pointer further describing the event(if needed).
+	//   - [err] error case the Watcher failed and wants to notify subscribers(see EventData).
+	Run(ctx context.Context) (bool, string, interface{}, error)
+}
+
+// Manager defines the interface between events management layer and the
+// core guest agent implementation.
+type Manager struct {
+	watchers    map[string]Watcher
+	subscribers map[string][]*eventSubscriber
+}
+
+// Config offers a mechanism for the consumers to configure the Manager behavior.
+type Config struct {
+	// Watchers lists the enabled watchers, of not provided all available watchers will be enabled.
+	Watchers []string
+}
+
+// EventData wraps the data communicated from a Watcher to a Subscriber.
+type EventData struct {
+	// Data points to the Watcher provided data.
+	Data interface{}
+	// Error is used when a Watcher has failed and wants communicate its subscribers about the error.
+	Error error
+}
+
+// EventCb defines the callback interface between watchers and subscribers. The arguments are:
+//   - evType a string defining the what event type triggered the call.
+//   - data a user context pointer to be consumed by the callback.
+//   - evData a event specific data pointer.
+//
+// The callback should return true if it wants to renew, returning false will case the callback
+// to be unregistered/unsubscribed.
+type EventCb func(evType string, data interface{}, evData *EventData) bool
+
+type eventSubscriber struct {
+	data interface{}
+	cb   EventCb
+}
+
+type eventBusData struct {
+	leaving bool
+	evType  string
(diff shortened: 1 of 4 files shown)


google_guest_agent/events/metadata/metadata.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+const (
+	// WatcherID is the metadata watcher's ID.
+	WatcherID = "metadata-watcher"
+)
+
+var (
+	// arbitrarily defined wait duration(keeps behavioral backward compatibility).
+	retryWaitDuration = 5 * time.Second
```
ChaitanyaKulkarni28, line 35: it doesn't seem to be used anywhere, was it intended to use on line 59?

google_guest_agent/events/metadata/metadata.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+			logger.Errorf("Error watching metadata: %s", err)
+			mp.failedPrevious = true
+		}
+	} else {
+		mp.failedPrevious = false
+	}
+
+	// If mds failed to respond we'll produce an event with a nil event data pointer.
+	return true, WatcherID + ",longpool", descriptor
```
ChaitanyaKulkarni28, line 79: if returned event ID (`WatcherID + ",longpool"`) is not expected to be unique we can define this as a constant. 
also the typo `longpoll`.

google_guest_agent/events/events.go
```diff
@@ -0,0 +1,221 @@
...
+	})
+	if err != nil {
+		logger.Errorf("Failed to initialize watchers: %+v", err)
+	}
+}
+
+// init initializes the known available event watchers.
+func initWatchers(watchers []Watcher) error {
+	for _, curr := range watchers {
+		// Ignore & log if we are accidentaly not properly setting the id.
```
ChaitanyaKulkarni28, line 84: Is this comment accurate? it says ignore and log but we're returning error?

dorileo: Post split leftover, will get it fixed. Thank you.

google_guest_agent/events/metadata/metadata.go
dorileo: Correct, line 59 will get updated. Thank you.

google_guest_agent/events/metadata/metadata.go
dorileo: Sounds good. Thanks.

google_guest_agent/events/events.go
```diff
@@ -0,0 +1,221 @@
...
+		defer wg.Done()
+
+		select {
+		case <-ctx.Done():
+			*leaving = true
+			channel <- &eventNotif{
+				leaving: true,
+			}
+		}
+	}(ctx, notifChan, waitGroup, &leaving)
```
ChaitanyaKulkarni28, line 164: do we need to pass `waitGroup` here, it will be already accessible inside go routine right?
Also, if we are passing it down to the function shouldn't it be the address (`&waitGroup`)?

dorileo: The intent was to make it explicit what is being used in each go routine (but obviously I failed to keep consistency).

google_guest_agent/events/metadata/metadata.go
dorileo: done.

google_guest_agent/events/events.go
dorileo: Done.

google_guest_agent/events/events.go
```diff
@@ -0,0 +1,221 @@
...
+
+	go func(channel <-chan *eventNotif, wg sync.WaitGroup) {
+		wg.Add(1)
+		defer wg.Done()
+
+		for {
```
ChaitanyaKulkarni28, line 170: I don't think we expect multiple cases here right? I guess we can simplify by adding range over channel, something like:

```
for notif := range channel {
...
}
```

dorileo: Correct, will get it changed. Thanks.

google_guest_agent/events/events.go
dorileo: I changed the `go func()` of all instances to receive only channels where we want to force/guard the channel direction. I removed all the other arguments.

google_guest_agent/events/events.go
dorileo: done. thanks.

google_guest_agent/events/events.go
```diff
@@ -0,0 +1,272 @@
...
+			return
+		case notif := <-syncBus:
+			if notif.leaving {
+				logger.Debugf("Got syncBus' leaving signal, leaving.")
+				return
+			}
+		}
+	}()
+
+	// Manages the the event processing avoiding blocking the watcher's go routines.
```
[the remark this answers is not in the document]
dorileo: Done.

google_guest_agent/events/events.go
```diff
@@ -0,0 +1,272 @@
...
+			}
+
+			keepMe := make([]*eventSubscriber, 0)
+			for _, curr := range mngr.subscribers[busData.evType] {
+				logger.Debugf("Running registered callback for event: %s", busData.evType)
+				renew := curr.cb(busData.evType, curr.data, busData.data)
+				if renew {
+					keepMe = append(keepMe, curr)
+				}
+				logger.Debugf("Returning from event \"%s\" subscribed callback, should renew?: %+v", busData.evType, renew)
```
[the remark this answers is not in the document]
dorileo: Done.

google_guest_agent/events/events.go
```diff
@@ -0,0 +1,272 @@
...
+			var evType string
+			var evData interface{}
+			var err error
+
+			defer wg.Done()
+
+			for renew := true; renew; {
+				renew, evType, evData, err = watcher.Run(ctx)
+
+				logger.Infof("Watcher(%s) returned event: %s, should renew?: %+v", watcher.ID(), evType, renew)
```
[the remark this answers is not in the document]
dorileo: done.

google_guest_agent/events/events.go
```diff
@@ -0,0 +1,272 @@
...
+
+// Watcher defines the interface between the events manager and the actual
+// watcher implementation.
+type Watcher interface {
+	// ID returns the watcher id.
+	ID() string
+	// Run implements the actuall "listening" strategy and emits a event "signal".
+	// It must return:
+	//   - [bool] if the watcher should renew(run again).
+	//   - [string] the event id.
```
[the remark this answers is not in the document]
dorileo: done.

PR #224, opened 2023-07-20
ssh_trusted_ca: enable the feature

dorileo:
Conditionally enable the event watcher and introduce the write end of the pipe implementation. Additionally we are introducing a new metadata client method to get one specific key (rather than the whole descriptor).

### DIFF
--- a/google_guest_agent/main.go
+++ b/google_guest_agent/main.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/events"
 	mdsEvent "github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/events/metadata"
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/events/sshtrustedca"
 	"github.com/GoogleCloudPlatform/guest-agent/metadata"
 	"github.com/GoogleCloudPlatform/guest-agent/utils"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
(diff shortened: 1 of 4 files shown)


metadata/metadata.go
```diff
@@ -300,3 +300,34 @@ func (c *Client) WriteGuestAttributes(ctx context.Context, key, value string) er
 	_, err = c.httpClient.Do(req)
 	return err
 }
+
+// GetKey gets a specific metadata key.
+func (c *Client) GetKey(key string) (string, error) {
```
ChaitanyaKulkarni28, line 305: Can we add context for HTTP calls?

metadata/metadata.go
```diff
@@ -300,3 +300,34 @@ func (c *Client) WriteGuestAttributes(ctx context.Context, key, value string) er
...
+	req, err := http.NewRequest("GET", c.metadataURL+key, nil)
+	if err != nil {
+		return "", err
+	}
+	req.Header.Add("Metadata-Flavor", "Google")
+
+	res, err := c.httpClient.Do(req)
+	if err == nil {
```
ChaitanyaKulkarni28, line 313: Is it really returning nil error in case of 404/412 error code, or do you mean to have `err != nil` condition?

metadata/metadata.go
```diff
@@ -300,3 +300,34 @@ func (c *Client) WriteGuestAttributes(ctx context.Context, key, value string) er
...
+		var statusCode string
+
+		switch res.StatusCode {
+		case 404:
+			statusCode = "(HTTP 404)"
+		case 412:
+			statusCode = "(HTTP 412)"
+		}
+
+		return "", fmt.Errorf("error connecting to metadata server%s: %+v", statusCode, err)
```
ChaitanyaKulkarni28, line 323: do we want to explicitly log all status codes? `statusCode` would be empty here if it's anything apart from 404/412?

google_guest_agent/main.go
```diff
@@ -170,12 +171,48 @@ func run(ctx context.Context) {
...
 	eventManager, err := events.New(eventsConfig)
 	if err != nil {
 		logger.Errorf("Error initializing event manager: %v", err)
 		return
 	}
 
+	eventManager.Subscribe(sshtrustedca.ReadEvent, nil, func(evType string, data interface{}, evData *events.EventData) bool {
+		// There was some error on the pipe watcher, just ignore it.
+		if evData.Error != nil {
+			return true
```
ChaitanyaKulkarni28, line 190: Do we need debug log here?

dorileo: Yep, Debug makes sense. Will do. Thanks.

metadata/metadata.go
dorileo: Sure thing. Thanks.

metadata/metadata.go
dorileo: Nope, I indeed meant `err == nil`,  in 404 and 412 error will be nil but with StatusCode set.

metadata/metadata.go
dorileo: Nice catch, will get it fixed. Thanks.

dorileo: Btw, we should propagate the error handling to the other places in the metadata package. I'm leaving a TODO to review it and come back to it later.

google_guest_agent/main.go
dorileo: done.

metadata/metadata.go
dorileo: Done.

metadata/metadata.go
dorileo: Done.

metadata/metadata.go
ChaitanyaKulkarni28: Got it, this should also check for non-nil error then before attempting to read response

dorileo: Sure. Will get it fixed. Thanks.

dorileo: Done.

PR #225, opened 2023-07-21
Support for reading & writing Root CA cert from UEFI variable

ChaitanyaKulkarni28:


google_guest_agent/agentcrypto/mtls_mds.go
```diff
@@ -0,0 +1,59 @@
+// Package agentcrypto provides various cryptography related utility functions for Guest Agent.
```
dorileo, line 1: Please add the license header.

google_guest_agent/agentcrypto/mtls_mds.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/uefi"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+const (
+	googleGUID                 = "8be4df61-93ca-11d2-aa0d-00e098032b8c"
```
dorileo, line 14: Can you please document each of these constants? Use the perspective of a system user (If I were to perform this steps manually what would mean each of these "artifacts"?).

google_guest_agent/uefi/uefi.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Variable represents UEFI Variable and its contents.
+type Variable struct {
+	Name       VariableName
+	Attributes []byte
+	Content    []byte
+}
+
+// VariablePath returns a path for UEFI variable on disk.
+func VariablePath(v VariableName) string {
```
dorileo, line 29: What about we transform this `VariablePath()` function into `func (v Variable) String()...` ?

google_guest_agent/uefi/uefi.go
```diff
@@ -0,0 +1,57 @@
...
+	if v.Name == "testname" && v.GUID == "testguid" {
```
dorileo, line 30: What about adding a `RootDir` field to `Variable`, if the `RootDir` is "" we use the default one, otherwise we assume it(or even some constructor to configure it, or any other option), that way we configure the `Variable.RootDir` with `t.TempDir()` in the unit tests, and remove this logic from this function?

google_guest_agent/uefi/uefi.go
```diff
@@ -0,0 +1,57 @@
...
+// Throws an error if variable is invalid or empty.
+func ReadVariable(v VariableName) (*Variable, error) {
+	path := VariablePath(v)
+	b, err := os.ReadFile(path)
+
+	if err != nil {
+		return nil, fmt.Errorf("error reading %q: %v", path, err)
+	}
+
+	// According to UEFI specification the first four bytes of the contents are attributes.
```
dorileo, line 47: This comment makes me wonder if for the mtls context we are using this four bytes to convey anything that's important for us (will change any behavior etc).

google_guest_agent/uefi/uefi_test.go
```diff
@@ -0,0 +1,64 @@
+package uefi
```
dorileo, line 1: Please add the license header.

dorileo (changes requested):

google_guest_agent/agentcrypto/certs.go
```diff
@@ -0,0 +1,21 @@
+package agentcrypto
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+)
+
+// VerifyCertificate validates certificate is in valid PEM format.
+func VerifyCertificate(cert []byte) error {
```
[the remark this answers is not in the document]
ChaitanyaKulkarni28: I want to keep general helper methods separate (more to come encrypt/decrypt) from the bootstrap workflow renamed it for better readability.

google_guest_agent/agentcrypto/mtls_mds.go
ChaitanyaKulkarni28: Done

google_guest_agent/uefi/uefi.go
ChaitanyaKulkarni28: Done

google_guest_agent/uefi/uefi.go
ChaitanyaKulkarni28: Done

google_guest_agent/uefi/uefi.go
ChaitanyaKulkarni28: These attributes are unrelated to content they are basically describing things like if variable is non-volatile/accessible at boot/runtime that won't change the actual content/behavior that we're interested in.

google_guest_agent/uefi/uefi_test.go
ChaitanyaKulkarni28: Done

dorileo (approved):

Lyllt8:
2023-07-21 12:11 GMT-06:00, Leandro Dorileo ***@***.***>:

PR #231, opened 2023-07-28
metadata: implement backoff strategy

dorileo:
Implement backoff strategy. Additionally moved all the http.Do() calls into a single method - reused for Get(), GetKey() and Watch().

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -31,12 +31,13 @@ import (
 const (
 	defaultMetadataURL = "http://169.254.169.254/computeMetadata/v1/"
 	defaultEtag        = "NONE"
-	metadataRecursive  = "/?recursive=true&alt=json"
-	metadataHang       = "&wait_for_change=true&timeout_sec=60"
+	defaultTimeout     = 60
 )
 
 var (
-	defaultTimeout = 70 * time.Second
+	// we backoff until 10s
+	backoffDuration = 100 * time.Millisecond
+	backoffAttempts = 100
 )
 
 // MDSClientInterface is the minimum required Metadata Server interface for Guest Agent.
@@ -45,6 +46,16 @@ type MDSClientInterface interface {
 	GetKey(context.Context, string) (string, error)
 	Watch(context.Context) (*Descriptor, error)
 	WriteGuestAttributes(context.Context, string, string) error
+}
+
+// requestConfig is used internally to configure an http request given its context.
+type requestConfig struct {
+	method     string
+	baseURL    string
+	hang       bool
+	recursive  bool
+	jsonOutput bool
+	timeout    int
 }
 
 // Client defines the public interface between the core guest agent and
@@ -61,7 +72,7 @@ func New() *Client {
 		metadataURL: defaultMetadataURL,
 		etag:        defaultEtag,
 		httpClient: &http.Client{
-			Timeout: defaultTimeout,
+			Timeout: defaultTimeout * time.Second,
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


metadata/metadata.go
```diff
@@ -309,30 +338,62 @@ func (c *Client) WriteGuestAttributes(ctx context.Context, key, value string) er
...
+	if ctx.Err() != nil {
+		return "", ctx.Err()
+	}
+
+	if cfg.hang {
+		c.updateEtag(resp)
+	}
+
+	defer resp.Body.Close()
+	md, err := ioutil.ReadAll(resp.Body)
```
ChaitanyaKulkarni28, line 396: I believe `ioutil` package is now deprecated. Since we're updating this code can we move to using the alternative `io.ReadAll`?

metadata/metadata.go
```diff
@@ -309,30 +338,62 @@ func (c *Client) WriteGuestAttributes(ctx context.Context, key, value string) er
...
+	if err != nil {
 		return "", fmt.Errorf("error connecting to metadata server: %+v", err)
 	}
 
-	defer res.Body.Close()
-	md, err := ioutil.ReadAll(res.Body)
+	statusCodeMsg := "error connecting to metadata server, status code: %d"
+	switch resp.StatusCode {
+	case 404, 412:
+		return "", fmt.Errorf(statusCodeMsg, resp.StatusCode)
```
ChaitanyaKulkarni28, line 384: should this be returning error for all non-2xx status codes?

metadata/metadata.go
```diff
@@ -47,6 +48,16 @@ type MDSClientInterface interface {
 	WriteGuestAttributes(context.Context, string, string) error
 }
 
+// requestConfig is used internally to configure an http request given its context.
+type requestConfig struct {
```
ChaitanyaKulkarni28, line 52: If we want to move all HTTP calls to MDS to a central location I think headers and body should also be included in requestConfig or you're thinking of keeping write calls separate?

metadata/metadata.go
```diff
@@ -31,12 +31,13 @@ import (
...
-	metadataRecursive  = "/?recursive=true&alt=json"
-	metadataHang       = "&wait_for_change=true&timeout_sec=60"
+	defaultTimeout     = 60
 )
 
 var (
-	defaultTimeout = 70 * time.Second
+	// we backoff until 10s
+	backoffDuration = 100 * time.Millisecond
+	backoffAttempts = 100
```
ChaitanyaKulkarni28, line 40: is this arbitrary number? 100 feels like a very big number for retries

dorileo: Totally arbitrary, I don't even intend to say these numbers make sense but just a small detail to add, the duration is in milliseconds.

metadata/metadata.go
dorileo: For now I don't plan to add write calls support to `requestConfig` or `client.do()`, at least not with this PR.

metadata/metadata.go
dorileo: it should, but I'll have to review all the "special codes" for now we are only handling this two, but others will follow.

metadata/metadata.go
dorileo: for sure. will do.

metadata/metadata.go
dorileo: for that purpose, I'm even dropping the `method` field.

dorileo: Done. Thanks.

PR #235, opened 2023-07-28
metadata event handler: updates and bug fixes

dorileo:
Metadata handler:
We shouldn't do anything (i.e. updating configuration) if the metadata passed in by the event watcher is nil.

Event Manager:
Fix an issue with event manager layer where the context and the sync channels were being handled as mutually exclusive causing the event manager to never "abort" after a cancel if the callbacks handler go routines are already setup and running.

Event Handler Test:
Added a test to exercise a call to cancel the context after the callback handlers go routines are already setup and running.

### DIFF
--- a/google_guest_agent/events/events.go
+++ b/google_guest_agent/events/events.go
@@ -80,9 +80,8 @@ type eventSubscriber struct {
 }
 
 type eventBusData struct {
-	leaving bool
-	evType  string
-	data    *EventData
+	evType string
+	data   *EventData
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


google_guest_agent/events/events.go
```diff
@@ -168,59 +167,76 @@ func (mngr *Manager) Run(ctx context.Context) {
...
-				continue
-			}
+				keepMe := make([]*eventSubscriber, 0)
+				for _, curr := range mngr.subscribers[busData.evType] {
+					logger.Debugf("Running registered callback for event: %s", busData.evType)
+					renew := curr.cb(busData.evType, curr.data, busData.data)
+					if renew {
+						keepMe = append(keepMe, curr)
+					}
+					logger.Debugf("Returning from event %q subscribed callback, should renew?: %+v", busData.evType, renew)
```
ChaitanyaKulkarni28, line 221: nit: `%t` for bool

dorileo: done. Thanks.

PR #238, opened 2023-07-31
metadata: don't try to use metadata before agentInit() is done

dorileo:
Avoid adding metadata dependent code before agentInit() is done and the routes are not certain to be ready. This avoid blocking too long before sending systemd-notify signal.

The logger then requires to have another attempt to configure the cloud logging client after we have metadata ready/available since the project name is a dependency to it and we use metadata to determine its value/name.

### DIFF
--- a/google_guest_agent/main.go
+++ b/google_guest_agent/main.go
@@ -130,16 +130,9 @@ func run(ctx context.Context) {
 		// Local logging is syslog; we will just use stdout in Linux.
 		opts.DisableLocalLogging = true
 	}
+
 	if os.Getenv("GUEST_AGENT_DEBUG") != "" {
 		opts.Debug = true
-	}
-
-	mdsClient = metadata.New()
-
-	var err error
-	newMetadata, err = mdsClient.Get(ctx)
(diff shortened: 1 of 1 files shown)


google_guest_agent/main.go
```diff
@@ -130,18 +130,11 @@ func run(ctx context.Context) {
...
 
-	mdsClient = metadata.New()
-
-	var err error
-	newMetadata, err = mdsClient.Get(ctx)
-	if err == nil {
-		opts.ProjectName = newMetadata.Project.ProjectID
-	}
-
 	if err := logger.Init(ctx, opts); err != nil {
```
Quintonamore, line 138: If err is declared here, why do we redeclare it again below on line 152?

Quintonamore (approved):

google_guest_agent/main.go
dorileo: in this case the `err` scope is the if block. We declare it on line 152 because the first user the line 153 is using a globally declared variable (a simply already declared variable would suffice).

PR #242, opened 2023-08-01
sshtrustedca watcher: fix concurrency error

dorileo:
Avoid trying to re-open the pipe for write until the write is finished. Additionally introduce a channel to communicate the context cancelation handling go routine to finish when returning to the event manager.

### DIFF
--- a/google_guest_agent/events/sshtrustedca/sshtrustedca_linux.go
+++ b/google_guest_agent/events/sshtrustedca/sshtrustedca_linux.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 	"syscall"
+	"time"
 
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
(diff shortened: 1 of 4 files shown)


google_guest_agent/events/sshtrustedca/sshtrustedca_linux.go
```diff
@@ -48,13 +49,30 @@ func createNamedPipe(pipePath string) error {
...
+// pipe is finised.
+func (mp *Watcher) finishedCb() {
+	mp.waitingWrite = false
+}
+
 // Run listens to ssh_trusted_ca's pipe open calls and report back the event.
 func (mp *Watcher) Run(ctx context.Context) (bool, string, interface{}, error) {
 	var canceled bool
 
+	for mp.waitingWrite {
```
Quintonamore, line 62: If we're expecting to use `waitingWrite` in multiple contexts, we should wrap this in a getter / setter that uses a mutex to protect reads / writes.

google_guest_agent/main.go
```diff
@@ -225,7 +225,12 @@ func run(ctx context.Context) {
 
 		// Make sure we close the pipe after we've done writing to it.
 		pipeData := evData.Data.(*sshtrustedca.PipeData)
-		defer pipeData.File.Close()
+		defer func() {
+			if err := pipeData.File.Close(); err != nil {
+				logger.Errorf("Failed to close pipe: %+v", err)
+			}
+			pipeData.Finished()
```
Quintonamore, line 232: I think we should try to keep this call in sshtreustedca's Run function, otherwise we're relying on a call to `Finished` that could never come leading to an infinite loop.

google_guest_agent/events/sshtrustedca/sshtrustedca_linux.go
dorileo: seems like a good call. will do.

google_guest_agent/main.go
dorileo: This is an "async call", we can't exactly call it from `Run()`, think of it as a bottom half strategy but with a sync up requirement.

google_guest_agent/events/sshtrustedca/sshtrustedca_linux.go
dorileo: Done. Thanks.

Quintonamore (approved):

PR #243, opened 2023-08-02
Add support for reading UEFI variables on windows

ChaitanyaKulkarni28:


google_guest_agent/uefi/uefi_windows.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// This call returns number of bytes written to the output buffer, unused, error
+	size, _, err := procGetFirmwareEnvironmentVariableW.Call(
+		uintptr(name),
+		uintptr(guid),
+		uintptr(unsafe.Pointer(&buffer[0])),
+		uintptr(uint32(len(buffer))),
+	)
+
+	if size == uintptr(0) {
```
dorileo, line 109: Is there a condition of read size > 0 and  `error`?

google_guest_agent/uefi/uefi_windows.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	return &Variable{
+		Name:       v,
+		Attributes: []byte{},
+		Content:    buffer[:size],
+	}, nil
+}
+
+// enablePrivilege enables the specified privilege for current process.
```
dorileo, line 120: are we changing the guest agent's process? is it applicable to the runtime environment of the guest-agent "service"?

google_guest_agent/uefi/uefi_windows.go
ChaitanyaKulkarni28: no, thing with these calls is that `error` is always non-nil we need to inspect primary return value. Here return value 0 indicates the failure.

google_guest_agent/uefi/uefi_windows.go
ChaitanyaKulkarni28: yep, verified on `windows2019`

PR #244, opened 2023-08-07
Add scheduler utility to run jobs at defined interval

ChaitanyaKulkarni28:


utils/scheduler/scheduler.go
```diff
@@ -0,0 +1,177 @@
...
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/agentcrypto"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	"github.com/robfig/cron/v3"
```
dorileo, line 26: what's the benefit of using `github.com/robfig/cron/v3` instead of implementing the scheduling interval using/handling a timer?

PR #246, opened 2023-08-09
sshca: transform the format from json to openssh

dorileo:
The return of the oslogin/certificates call returns a json formated data structure and openssh waits a list of certificates (one per line). This change introduces such transformation.

### DIFF
--- a/google_guest_agent/main.go
+++ b/google_guest_agent/main.go
@@ -18,6 +18,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


google_guest_agent/main.go
```diff
@@ -39,6 +40,16 @@ import (
 	"github.com/go-ini/ini"
 )
 
+// Certificates wrapps a list of certificate authorities.
```
Quintonamore, line 43:
```suggestion
// Certificates wraps a list of certificate authorities.
```

Small typo.

Quintonamore (approved):

PR #248, opened 2023-08-09
Run telemetry every 24 hours, record pretty name on linux

adjackura:


google_guest_agent/main.go
```diff
@@ -119,6 +119,28 @@ func runUpdate(ctx context.Context) {
...
+				ShortName:     osInfo.OS,
+				Version:       osInfo.VersionID,
+				KernelRelease: osInfo.KernelRelease,
+				KernelVersion: osInfo.KernelVersion,
+			}
+			if err := telemetry.Record(ctx, mdsClient, d); err != nil {
+				logger.Debugf("Error recording telemetry: %v", err)
+			}
+		}
+		time.Sleep(24 * time.Hour)
```
dorileo, line 140: Nope, please, please use the [just introduced scheduler](https://github.com/GoogleCloudPlatform/guest-agent/tree/main/utils/scheduler)

dntczdx: Hi Leo, your link looks like a 404 not found

ChaitanyaKulkarni28: Please use this [updated link](https://github.com/GoogleCloudPlatform/guest-agent/tree/main/google_guest_agent/scheduler)

dntczdx:
We can also flip the enabling flag since the functionality has been verified.

google_guest_agent/main.go
adjackura: I can update this to use that but I'm a little concerned with that library in general, calling Get has unexpected side effects, and you are importing a library I don't think you really need. 
Writing scheduling logic should be pretty straightforward, especially since I don't think you need any of the complexities of that library? Also you probably should have your main configure your scheduler and not have the idea of default jobs, that can easily cause issues for you.

google_guest_agent/telemetry/telemetry.go
```diff
@@ -81,6 +89,64 @@ func Record(ctx context.Context, client metadata.MDSClientInterface, d Data) err
 		"X-Google-Guest-Agent": formatGuestAgent(d),
 		"X-Google-Guest-OS":    formatGuestOS(d),
 	}
-	_, err := client.GetKey(ctx, "project", headers)
```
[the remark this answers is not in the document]
adjackura: We found that this caused a double lookup on the metadata server side (this appears to all be server logic, nothing the client was doing), apparently just using the root node solves this. We theorize that the server is doing two lookups, one for the parent and one for the key.

google_guest_agent/telemetry/telemetry.go
```diff
@@ -81,6 +89,64 @@ func Record(ctx context.Context, client metadata.MDSClientInterface, d Data) err
...
+}
+
+// ShouldEnable returns true as long as DisableTelemetry is not set in metadata.
+func (j *Job) ShouldEnable(ctx context.Context) bool {
+	md, err := j.client.Get(ctx)
+	if err != nil {
+		return false
+	}
+
+	return !md.Instance.Attributes.DisableTelemetry && !md.Project.Attributes.DisableTelemetry
```
dntczdx, line 151: I think we can flip the logic to default=TRUE now since we have verified the  behaviors on prod and the launch is approved

google_guest_agent/telemetry/telemetry.go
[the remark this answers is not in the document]
dorileo: Can you add a comment line describing this behavior?

Thanks.

google_guest_agent/telemetry/telemetry.go
adjackura: Lets get that into its own PR, as soon as this is Ill send another to change that.

google_guest_agent/telemetry/telemetry.go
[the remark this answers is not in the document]
adjackura: udpated

PR #250, opened 2023-08-10
events: multiplex event watchers

dorileo:
Allow event watchers to emit more than a single event signal. That way the same watcher can have "specialized" event types, i.e.: metadata watcher could have a "first-pool" event notification sent when we have established that we can start querying it.

### DIFF
--- a/google_guest_agent/events/events.go
+++ b/google_guest_agent/events/events.go
@@ -35,13 +35,14 @@ var (
 type Watcher interface {
 	// ID returns the watcher id.
 	ID() string
+	// Events return a slice with all the event types a given Watcher handles.
+	Events() []string
 	// Run implements the actuall "listening" strategy and emits a event "signal".
 	// It must return:
 	//   - [bool] if the watcher should renew(run again).
-	//   - [string] the event type id.
 	//   - [interface{}] a event context data pointer further describing the event(if needed).
 	//   - [err] error case the Watcher failed and wants to notify subscribers(see EventData).
-	Run(ctx context.Context) (bool, string, interface{}, error)
+	Run(ctx context.Context, evType string) (bool, interface{}, error)
 }
 
 // Manager defines the interface between events management layer and the
@@ -63,6 +64,14 @@ type EventData struct {
 	Data interface{}
 	// Error is used when a Watcher has failed and wants communicate its subscribers about the error.
 	Error error
+}
+
+// WatcherEventType wraps/couples together a Watcher and an event type.
+type WatcherEventType struct {
+	// watcher is the watcher implementation for a given event type.
+	watcher Watcher
+	// evType idenfities the event type this object refences to.
+	evType string
 }
 
 // EventCb defines the callback interface between watchers and subscribers. The arguments are:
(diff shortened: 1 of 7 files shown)


google_guest_agent/events/events_test.go
```diff
@@ -72,13 +72,19 @@ func (tprod *testWatcher) ID() string {
 	return tprod.watcherID
 }
 
-func (tprod *testWatcher) Run(ctx context.Context) (bool, string, interface{}, error) {
-	if tprod.counter >= tprod.maxCount {
-		return false, "", nil, nil
-	}
+func (tprod *testWatcher) Events() []string {
+	return []string{tprod.watcherID + ",test-event"}
```
Quintonamore, line 76: Nit: For slightly more completeness, I think testing an `Events` function with more than one event in its slice would be useful.

Quintonamore (approved):

google_guest_agent/events/events_test.go
dorileo: Yep, indeed.

dorileo: Done. thanks.

google_guest_agent/events/events.go
```diff
@@ -248,19 +267,18 @@ func (mngr *Manager) Run(ctx context.Context) {
 
 	// Creates a goroutine for each registered watcher and keep handling its
```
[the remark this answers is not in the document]
dorileo: Done, thanks.

google_guest_agent/events/metadata/metadata.go
```diff
@@ -55,8 +55,13 @@ func (mp *Watcher) ID() string {
...
 }
 
+// Events returns an slice with all implemented events.
+func (mp *Watcher) Events() []string {
+	return []string{LongpollEvent}
+}
+
 // Run listens to metadata changes and report back the event.
-func (mp *Watcher) Run(ctx context.Context) (bool, string, interface{}, error) {
+func (mp *Watcher) Run(ctx context.Context, evType string) (bool, interface{}, error) {
```
ChaitanyaKulkarni28, line 64: I know right now there's only one event type but do we also want to refactor `Run()` to execute based on the received `evType`? Or that's for future whenever watcher introduces new event type?

dorileo: Do you mean in the Watcher implementations?

dorileo: Oh yeah! you are quoting the watcher implementation...

ChaitanyaKulkarni28:
Not directly related to this PR but while we're updating event manager, do you also want to enforce singleton pattern for it?

google_guest_agent/events/metadata/metadata.go
dorileo: Yep, for sure, so if we want to support multiple events the Run() implementation will decide what event to handle depending on the receiving evType... for metadata for example we can add a "first-metadata-ready" and handle it differently ....

dorileo:
@ChaitanyaKulkarni28 I'll do the singleton interface refactor in a followup PR.

google_guest_agent/events/metadata/metadata.go
ChaitanyaKulkarni28: Right, I meant do you intend to refactor it in this PR or later when we introduce more events? 

Basically have a switch case or something and `Run()` executes only on `LongpollEvent evType` otherwise throw an error or no-op.

PR #253, opened 2023-08-11
Enable telemetry collection by default

adjackura:
Only commit once docs are updated.


dntczdx (commented):
modify the README along with this enabling?

dntczdx (approved):

PR #255, opened 2023-08-11
Support for store windows certs in certstore

ChaitanyaKulkarni28:
- This adds support for generating PFX files and storing certificates in Windows Certstore


google_guest_agent/agentcrypto/crypto_util.go
```diff
@@ -38,6 +39,21 @@ func parseCertificate(cert []byte) (*x509.Certificate, error) {
 	return x509Cert, nil
 }
 
+// parsePvtKey validates the key is in valid format and returns the EC Private Key.
+func parsePvtKey(pemKey []byte) (*ecdsa.PrivateKey, error) {
+	key, _ := pem.Decode(pemKey)
```
dorileo, line 44: any trick in the `Decode()` error return here?

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	golang.org/x/sys v0.11.0
 	google.golang.org/grpc v1.57.0
 	google.golang.org/protobuf v1.31.0
+	software.sslmate.com/src/go-pkcs12 v0.2.1
```
dorileo, line 22: I guess we'll need to update the third party license again, right?

google_guest_agent/agentcrypto/mtls_mds.go
```diff
@@ -208,3 +207,16 @@ func (j *CredsJob) ShouldEnable(ctx context.Context) bool {
 	_, err := j.client.GetKey(ctx, clientCertsKey, nil)
 	return err == nil
 }
+
+func writeToFile(content []byte, outputFile string) error {
```
dorileo, line 211: we might have a few duplicates functions like this in the code base.

google_guest_agent/agentcrypto/mtls_mds_windows.go
```diff
@@ -0,0 +1,201 @@
...
+		return fmt.Errorf("failed to get cert context for PFX from store: %w", err)
+	}
+	defer windows.CertFreeCertificateContext(crtCtx)
+
+	// Add certificate to personal store.
+	if err := addCtxToLocalSystemStore(my, crtCtx, uint32(windows.CERT_STORE_ADD_NEWER)); err != nil {
+		return fmt.Errorf("failed to store pfx cert context: %w", err)
+	}
+
+	if err := deleteCert(prevCtx, my); err != nil {
```
dorileo, line 121: What if we restart the agent we don't know "the previous context"?

go.mod
ChaitanyaKulkarni28: yes, I am planning to have a follow up PR for updating the licenses after this.

google_guest_agent/agentcrypto/crypto_util.go
ChaitanyaKulkarni28: `Decode()` does not return error and we don't expect/care any remaining bytes so its ignored.

google_guest_agent/agentcrypto/mtls_mds.go
ChaitanyaKulkarni28: moved this to utils.

google_guest_agent/agentcrypto/mtls_mds_windows.go
ChaitanyaKulkarni28: Added comments to summarize the behavior.

google_guest_agent/agentcrypto/mtls_mds.go
```diff
@@ -131,62 +120,71 @@ func (j *CredsJob) extractKey(importBlob *tpm.ImportBlob) ([]byte, error) {
...
+// Invoke-RestMethod -Uri https://169.254.169.254 -Method Get -Headers @{"Metadata-Flavor"="Google"} -Certificate $cert
 func (j *CredsJob) Run(ctx context.Context) (bool, error) {
-	// defaultCredsDir is the directory location for MTLS MDS credentials.
-	defaultCredsDir := "/etc/pki/tls/certs/mds"
+	logger.Infof("Fetching Root CA cert...")
 
-	// TODO: Finalize on where to store certificates on windows.
-	if runtime.GOOS == "windows" {
-		defaultCredsDir = `C:\Users`
+	v, err := j.readRootCACert(googleRootCACertUEFIVar)
```
dorileo, line 170: Can you add a EUFI variable check in the begging of this function and if the variable is not set we just return? That way we return early and use that as a signal that the feature is not enable?

dorileo: You can do it in a follow-up PR though.

dorileo:
Overall /lgtm

PR #257, opened 2023-08-14
Add pkcs12 license and update retry logic

ChaitanyaKulkarni28:
Add pkcs12 license and update retry logic to not retry on non-retriable errors.

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -20,7 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


metadata/metadata.go
```diff
@@ -273,12 +273,26 @@ func (c *Client) updateEtag(resp *http.Response) bool {
...
+}
+
 func (c *Client) retry(ctx context.Context, cfg requestConfig) (string, error) {
 	for i := 1; i <= backoffAttempts; i++ {
 		resp, err := c.do(ctx, cfg)
 
-		// If the context was canceled just return the error and don't retry.
-		if err != nil && errors.Is(err, context.Canceled) {
+		// Check if error is retriable, if not just return the error and don't retry.
+		if err != nil && !shouldRetry(resp, err) {
```
dorileo, line 295: It should fix one of the issues I noticed, but additionally we want ScheduleJob() to not block the caller, it all should run in its own go routine - with this new non retry'able logic we'll see the symptom go away but we may stumble on it again later when the ShouldEnable() depends on a long lasting operation. We can address it in a follow up PR though.

metadata/metadata.go
```diff
@@ -289,7 +303,15 @@ func (c *Client) retry(ctx context.Context, cfg requestConfig) (string, error) {
 			continue
 		}
 
-		return resp, nil
+		defer resp.Body.Close()
+		md, err := io.ReadAll(resp.Body)
+		if err != nil {
+			logger.Errorf("Failed to read metadata server response bytes: %+v", err)
```
dorileo, line 309: Since we are here, can you turn this Errorf into a Debugf? And compute a summary to display on the last line?

ChaitanyaKulkarni28: done

metadata/metadata.go
ChaitanyaKulkarni28: Considering the purpose of `ShouldEnable()` it should not be a long running operation to have a noticeable block. Updated to run as separate go routine and not block agent main.

PR #263, opened 2023-08-24
Find previous cert for cleanup using one stored on disk

ChaitanyaKulkarni28:
* Add support for finding a certificate based on issuer name and serial number
* Use that to lookup the certificate stored on disk for cleaning up the previous certificate from certstore

### DIFF
--- a/google_guest_agent/agentcrypto/mtls_mds_windows.go
+++ b/google_guest_agent/agentcrypto/mtls_mds_windows.go
@@ -41,6 +41,11 @@ const (
 	my = "MY"
 	// root is predefined cert store for root trusted CA certs.
 	root = "ROOT"
+	// certificateIssuer is the issuer of client/root certificates for MDS mTLS.
+	certificateIssuer = "google.internal"
+	// maxCertEnumeration specifies the maximum number of times to search for a certificate
+	// with a serial number from a given issuer before giving up.
+	maxCertEnumeration = 5
 )
 
 var (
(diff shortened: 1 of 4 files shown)


google_guest_agent/agentcrypto/mtls_mds_windows.go
```diff
@@ -77,8 +79,63 @@ func (j *CredsJob) writeRootCACert(cacert []byte, outputFile string) error {
...
+		0,
+		windows.CERT_SYSTEM_STORE_LOCAL_MACHINE,
+		uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(storeName))))
+	if err != nil {
+		return nil, fmt.Errorf("failed to open cert store: %w", err)
+	}
+
+	var prev *windows.CertContext
+
+	for i := 1; i <= 5; i++ {
```
dorileo, line 98: Is this a retry?

google_guest_agent/agentcrypto/mtls_mds_windows.go
```diff
@@ -119,22 +176,30 @@ func (j *CredsJob) writeClientCredentials(creds []byte, outputFile string) error
...
 
-	// Best effort to cleanup previous certificates and keep only the latest.
-	// When the agent restarts, it will not have the previous context in memory.
-	// Therefore, it will not remove any old certificates that are not required.
-	// If there are multiple similar certificates, clients should try to use the one
-	// with the longest expiry date, as that will be the latest version of it.
-	// TODO: See if we can have a more robust process for cleaning up old certificates,
-	// while ensuring that we don't delete any that were not installed by the agent process.
+	// Search for previous certificate if its not already in memory.
+	if prevCtx == nil && num != "" {
```
dorileo, line 180: Knowing if we could get the previously added cert in the certstore is a strong requirement - I'd say that cleanup + addition are atomic operations - I'd suggest moving the "find certificate" to the very beginning, with that if we fail to find it we can abort before we've written the new certificate, that way we can try again the whole process without dealing with inconsistent state.

dorileo: We talked offline, it seems to be fine.

google_guest_agent/agentcrypto/mtls_mds_windows.go
ChaitanyaKulkarni28: added comments to explain the behavior

PR #265, opened 2023-08-28
run: split command running utilities into its own package

dorileo:
Move it from main to its own package and add unit tests for them.

### DIFF
--- a/google_guest_agent/run/run.go
+++ b/google_guest_agent/run/run.go
@@ -1,0 +1,97 @@
+//  Copyright 2023 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package run is a package with utilities for running command and handling results.
+package run
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+// Result wraps a command execution result.
+type Result struct {
+	// Exit code. Set to -1 if we failed to run the command.
+	ExitCode int
+	// Stderr or err.Error if we failed to run the command.
+	StdErr string
+	// Stdout or "" if we failed to run the command.
+	StdOut string
+}
+
+// Error return an error containing the stderr content.
+func (e Result) Error() string {
+	return strings.TrimSuffix(e.StdErr, "\n")
+}
+
+// Quiet runs a command and doesn't return a result, but an error in case of failure.
+func Quiet(name string, args ...string) error {
+	res := execCommand(exec.Command(name, args...))
+	if res.ExitCode != 0 {
+		return res
+	}
+	return nil
+}
+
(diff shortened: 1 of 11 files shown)


google_guest_agent/run/run_test.go
```diff
@@ -0,0 +1,188 @@
...
+	"testing"
+	"time"
+)
+
+// builds a set of data to be used in the tests
+func buildDataContent(t *testing.T) (string, error) {
+	t.Helper()
+	rootDir := path.Join(t.TempDir(), fmt.Sprintf("%d", time.Now().UnixNano()))
+	if err := Quiet("mkdir", "-p", rootDir); err != nil {
+		return rootDir, err
```
ChaitanyaKulkarni28, line 32: Since `t` is already passed here we can do `t.Fatalf` here itself instead of returning and checking for error again at callers.
Also, since this is helper method and also used for testing `Quiet()`, I think we should not use it here and rather rely on standard library for that?

google_guest_agent/run/run_test.go
```diff
@@ -0,0 +1,188 @@
...
+	tests := []string{
+		"grep -R datax " + testDir,
+		"rm -R /root/data",
+	}
+
+	for _, curr := range tests {
+		t.Run(curr, func(t *testing.T) {
+			tokens := strings.Split(curr, " ")
+			if err := Quiet(tokens[0], tokens[1:]...); err == nil {
+				t.Errorf("command succeed, expected failure.")
```
ChaitanyaKulkarni28, line 80: Can we update the failure messages to include input and identify the function?

google_guest_agent/run/run.go
```diff
@@ -0,0 +1,97 @@
...
+	// Stdout or "" if we failed to run the command.
+	StdOut string
+}
+
+// Error return an error containing the stderr content.
+func (e Result) Error() string {
+	return strings.TrimSuffix(e.StdErr, "\n")
+}
+
+// Quiet runs a command and doesn't return any status.
```
ChaitanyaKulkarni28, line 44: is this comment accurate? we're returning `res` on error right?

google_guest_agent/run/run_test.go
dorileo: Yes and Yes. I'll get it done. Thanks.

google_guest_agent/run/run.go
dorileo: Ah! I meant to change this comment, it was wrong before me, it meant it doesn't return status but returns error. I'll get it changed. Thanks.

google_guest_agent/run/run_test.go
dorileo: The input is identified in the `t.Run()` name, but it's harmful to repeat in the error msg. Will get it changed. Thanks.

google_guest_agent/run/run_test.go
dorileo: Done. Thanks.

google_guest_agent/run/run.go
dorileo: Done. Thanks.

google_guest_agent/run/run_test.go
dorileo: done. Thanks.

google_guest_agent/run/run_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"strings"
+	"testing"
+	"time"
+)
+
+// builds a set of data to be used in the tests
+func buildDataContent(t *testing.T) string {
+	t.Helper()
+	rootDir := path.Join(t.TempDir(), fmt.Sprintf("%d", time.Now().UnixNano()))
+	if err := exec.Command("mkdir", "-p", rootDir).Run(); err != nil {
```
ChaitanyaKulkarni28, line 32: why not just use `os.MkdirAll()` instead of exec?

dorileo: Of course.

dorileo: Done. Thanks.

PR #268, opened 2023-08-29
run: change all commands to use CommandContext

dorileo:
With that we can prevent commands taking longer than expected when we are handling context cancellation (i.e. SIGTERM, SIGINT etc).

### DIFF
--- a/google_guest_agent/diagnostics.go
+++ b/google_guest_agent/diagnostics.go
@@ -17,11 +17,11 @@ package main
 import (
 	"context"
 	"encoding/json"
-	"os/exec"
 	"reflect"
 	"strconv"
 	"sync/atomic"
 
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
 	"github.com/GoogleCloudPlatform/guest-agent/utils"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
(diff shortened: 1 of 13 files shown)


google_guest_agent/diagnostics.go
```diff
@@ -116,13 +116,14 @@ func (d *diagnosticsMgr) set(ctx context.Context) error {
...
+
 	go func() {
 		logger.Infof("Diagnostics: collecting logs from the system.")
-		out, err := cmd.CombinedOutput()
-		logger.Infof(string(out[:]))
-		if err != nil {
-			logger.Infof("Error collecting logs: %v", err)
+		res := run.WithOutput(ctx, diagnosticsCmd, args...)
+		logger.Infof(res.StdOut)
+		logger.Infof(res.StdErr)
```
ChaitanyaKulkarni28, line 124: should we log `res.StdErr` only if `ExitCode != 0`?

dorileo: Previous code was using combined output so out is both stdout and stderr, I guess we want to expose a function to wrap combined as well and use it here - and be "more compatible" with the current implementation.

dorileo: done.

google_guest_agent/run/run.go
```diff
@@ -68,6 +70,23 @@ func WithOutputTimeout(ctx context.Context, timeout time.Duration, name string,
...
 }
 
+// WithCombinedOutput returns a result with stderr and stdout combined in the Combined
+// member of Result.
+func WithCombinedOutput(ctx context.Context, name string, args ...string) *Result {
+	cmd := exec.CommandContext(ctx, name, args...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return &Result{
+			ExitCode: -1,
```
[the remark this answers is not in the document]
dorileo: done. Thanks.

google_guest_agent/diagnostics.go
```diff
@@ -116,13 +116,13 @@ func (d *diagnosticsMgr) set(ctx context.Context) error {
...
 	go func() {
 		logger.Infof("Diagnostics: collecting logs from the system.")
-		out, err := cmd.CombinedOutput()
-		logger.Infof(string(out[:]))
-		if err != nil {
-			logger.Infof("Error collecting logs: %v", err)
+		res := run.WithCombinedOutput(ctx, diagnosticsCmd, args...)
+		logger.Infof(res.Combined)
+		if res.ExitCode != 0 {
+			logger.Infof("Error collecting logs: %v", err.Error())
```
ChaitanyaKulkarni28, line 125: sorry missed this while reviewing earlier. I believe instead of `err.Error()` this should be `res. StdErr` now? There's no `err` in this go routine now.

Also, I see it was logged at `info` level before but considering its an error do we want to change it to `warn`?

dorileo: Good catch, it should be `res.Error()` actually. Sure I'll change the logging to warn.

dorileo: done. Thanks.

PR #271, opened 2023-09-05
restorecon: first try to determine if it's installed

dorileo:
Not all distributions will have restorecon installed by default, rather than running the command and waiting for it to fail we better check if it it's available before running the command.

### DIFF
--- a/google_guest_agent/instance_setup.go
+++ b/google_guest_agent/instance_setup.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
 	"runtime"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


google_guest_agent/instance_setup.go
```diff
@@ -265,7 +266,10 @@ func generateSSHKeys(ctx context.Context) error {
 			logger.Warningf("Generated key is malformed, not uploading")
 		}
 	}
-	run.Quiet(ctx, "restorecon", "-FR", hostKeyDir)
+	_, err = exec.LookPath("restorecon")
+	if err == nil {
+		run.Quiet(ctx, "restorecon", "-FR", hostKeyDir)
```
ChaitanyaKulkarni28, line 271: i think we should at-least log error if any

dorileo: Nope, it's actually an error condition. I'll make it return an error.

dorileo: done. thanks.

PR #273, opened 2023-09-06
service: remove the use of the service library

dorileo:


ChaitanyaKulkarni28:
Can you also remove the license - `THIRD_PARTY_LICENSES/github.com/kardianos/service/LICENSE`

google_guest_agent/events/metadata/metadata.go
```diff
@@ -71,6 +66,12 @@ func (mp *Watcher) Run(ctx context.Context, evType string) (bool, interface{}, e
 					logger.Errorf("Network error when requesting metadata, make sure your instance has an active network and can reach the metadata server.")
 				}
 			}
+
+			if errors.Is(err, context.Canceled) {
+				logger.Infof("Metadata watcher: context was canceled, preparing to leave...")
+				return false, descriptor, nil
```
ChaitanyaKulkarni28, line 72: should this return an `err` instead of `nil`

google_guest_agent/main.go
```diff
@@ -144,8 +146,7 @@ func runAgent(ctx context.Context) {
 	}
 
 	if err := logger.Init(ctx, opts); err != nil {
-		fmt.Printf("Error initializing logger: %v", err)
-		os.Exit(1)
+		return fmt.Errorf("Error initializing logger: %v", err)
```
ChaitanyaKulkarni28, line 149: nit: error strings here and below should not be capitalized

google_guest_agent/main.go
```diff
@@ -262,21 +263,18 @@ func closer(c io.Closer) {
...
-	if action == "noservice" {
-		runAgent(ctx)
-		os.Exit(0)
-	}
+	ctx, cancelContext := context.WithCancel(context.Background())
+	sigBus := make(chan os.Signal, 1)
+
+	signal.Notify(sigBus, syscall.SIGTERM, syscall.SIGINT)
+	go func() {
+		select {
```
ChaitanyaKulkarni28, line 271: Since we don't expect multiple cases here we can simplify and log signal as - 
```
go func() {
        signal := <-sigBus
        // log agent is exiting because of signal it received
	cancelContext()
}()
```

google_guest_agent/main.go
```diff
@@ -262,21 +263,18 @@ func closer(c io.Closer) {
...
-	}
-
-	if action == "noservice" {
-		runAgent(ctx)
-		os.Exit(0)
-	}
+	ctx, cancelContext := context.WithCancel(context.Background())
+	sigBus := make(chan os.Signal, 1)
+
+	signal.Notify(sigBus, syscall.SIGTERM, syscall.SIGINT)
```
ChaitanyaKulkarni28, line 269: Do we also need to listen any other signals like SIGHUP/SIGABRT?

google_guest_agent/events/metadata/metadata.go
dorileo: Not really, it's not an error condition, it's handling the context cancelation, meaning that we should only greacefuly leave and not try to renew (hence the first return value as false).

google_guest_agent/main.go
dorileo: Indeed. Thanks.

google_guest_agent/main.go
dorileo: Indeed thanks.

google_guest_agent/main.go
dorileo: `SIGHUP` doesn't make sense in the context, `SIGABRT` will not be the case as systemd will never do an `abort()` and the signal expects the process to finish ASAP so not much of anything to expect from a handler. `SIGQUIT` however, makes sense as it's a similar signal to SIGTERM but with a slightly different condition. So yep, adding `SIGQUIT`.

google_guest_agent/main.go
dorileo: Done. Thanks.

google_guest_agent/main.go
dorileo: Done. Thanks.

google_guest_agent/main.go
dorileo: Done. Thanks.

PR #274, opened 2023-09-06
metadata watcher: introduce ReadyEvent

dorileo:
With this event we can have do all the metadata dependent setup after we get the event generated by the watcher.

### DIFF
--- a/google_guest_agent/events/metadata/metadata.go
+++ b/google_guest_agent/events/metadata/metadata.go
@@ -17,6 +17,7 @@ package metadata
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/url"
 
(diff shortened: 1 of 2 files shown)


google_guest_agent/events/metadata/metadata.go
```diff
@@ -52,11 +64,42 @@ func (mp *Watcher) ID() string {
...
-	return []string{LongpollEvent}
+	return []string{ReadyEvent, LongpollEvent}
 }
 
-// Run listens to metadata changes and report back the event.
-func (mp *Watcher) Run(ctx context.Context, evType string) (bool, interface{}, error) {
+func (mp *Watcher) runReadyWatcher(ctx context.Context, evType string) (bool, interface{}, error) {
+	descriptor, err := mp.client.Get(ctx)
+	if err != nil {
+		return false, nil, fmt.Errorf("Failed to get metadata descriptor: %+v", err)
```
Quintonamore, line 73: Should we have similar logging logic as with `runLongpollWatcher` here? My only concern is `failedPrevious` is only being used one of the run functions instead of both.

This isn't a requirement, but maybe error logging and handling should happen outside of both `runReadyWatcher` and `runLongpollWatcher`.

dorileo: It's a one shot only, we are not retrying (metadata package does) so we'll not exactly running the risk of having duplicated log messages. If we decide to implement a retry in this level then sure we should.

dorileo:
We have better solutions to this in the pipe, closing this one.

PR #279, opened 2023-09-12
metadata: avoid issues if MDS has returned empty response body

dorileo:
If MDS returns an empty response body we'll not be able to unmarshal it resulting in a null pointer being passed down to metadata package users.


dorileo:
Doesn't seem correct or necessary; Killing it for now.

PR #284, opened 2023-09-18
metadata: use url package to form metadata URLs

dorileo:
By using the url package to form metadata URLs we avoid having the weirdly formed URLs i.e. "v1//?" ...

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


metadata/metadata.go
```diff
@@ -386,35 +390,34 @@ func (c *Client) WriteGuestAttributes(ctx context.Context, key, value string) er
...
 	if cfg.hang {
-		urlTokens = append(urlTokens, "wait_for_change=true")
-		urlTokens = append(urlTokens, "last_etag="+c.etag)
+		values.Add("wait_for_change", "true")
+		values.Add("last_etag", c.etag)
 	}
 
 	if cfg.timeout > 0 {
-		urlTokens = append(urlTokens, fmt.Sprintf("timeout_sec=%d", cfg.timeout))
+		values.Add("timeout_sec", fmt.Sprintf("%d", cfg.timeout))
```
ChaitanyaKulkarni28, line 406: isn't `timeout` already an int?

metadata/metadata.go
```diff
@@ -318,13 +319,9 @@ func (c *Client) retry(ctx context.Context, cfg requestConfig) (string, error) {
...
-	// url.JoinPath first exists in go 1.19
-	key = strings.TrimPrefix(key, "/")
-	var reqURL string
-	if strings.HasSuffix(c.metadataURL, "/") {
-		reqURL = c.metadataURL + key
-	} else {
-		reqURL = c.metadataURL + "/" + key
+	reqURL, err := url.JoinPath(c.metadataURL, key)
+	if err != nil {
+		return "", fmt.Errorf("Failed to form metadata url: %+v", err)
```
ChaitanyaKulkarni28, line 324: nit: error strings should not be capitalized. Here and below error return messages

metadata/metadata.go
dorileo: Yep, it is. `values.Add()` is `values.Add(string, string)` that's why I'm converting it to string.

metadata/metadata.go
dorileo: Done. Thanks.

PR #285, opened 2023-09-18
oslogin: add PAMless Authorization Stack configuration

dorileo:
Introduces the configuration for a simplified authorization stack where
PAM modules are used only for 2fa. For now it's not widely available
and is protected behind Unstable.pamless_auth_stack configuration.

### DIFF
--- a/google_guest_agent/oslogin.go
+++ b/google_guest_agent/oslogin.go
@@ -99,7 +99,12 @@ func (o *osloginMgr) set(ctx context.Context) error {
 		logger.Infof("Disabling OS Login")
 	}
 
-	if err := writeSSHConfig(enable, twofactor, skey); err != nil {
+	// [Unstable] configuration section has no long term stability or support guarantees/promises.
+	// Configurations defined in the Unstable section will be by default disabled and is intended to
+	// isolate under development features.
+	pamlessAuthStack := config.Section("Unstable").Key("pamless_auth_stack").MustBool(false)
+
+	if err := writeSSHConfig(enable, twofactor, pamlessAuthStack, skey); err != nil {
 		logger.Errorf("Error updating SSH config: %v.", err)
 	}
 
@@ -107,7 +112,7 @@ func (o *osloginMgr) set(ctx context.Context) error {
 		logger.Errorf("Error updating NSS config: %v.", err)
 	}
 
-	if err := writePAMConfig(enable, twofactor); err != nil {
+	if err := writePAMConfig(enable, twofactor, pamlessAuthStack); err != nil {
 		logger.Errorf("Error updating PAM config: %v.", err)
 	}
 
@@ -187,7 +192,7 @@ func writeConfigFile(path, contents string) error {
 	return nil
 }
 
-func updateSSHConfig(sshConfig string, enable, twofactor, skey bool) string {
+func updateSSHConfig(sshConfig string, enable, twofactor, pamlessAuthStack, skey bool) string {
 	// TODO: this feels like a case for a text/template
 	challengeResponseEnable := "ChallengeResponseAuthentication yes"
 	authorizedKeysCommand := "AuthorizedKeysCommand /usr/bin/google_authorized_keys"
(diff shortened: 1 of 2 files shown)


dorileo:
Note: I'll introduce a mechanism to keep backward compatibility until the feature goes GA.

google_guest_agent/oslogin.go
```diff
@@ -99,15 +99,18 @@ func (o *osloginMgr) set(ctx context.Context) error {
 		logger.Infof("Disabling OS Login")
 	}
 
-	if err := writeSSHConfig(enable, twofactor, skey); err != nil {
+	// Unstable configuration section is a section with no guarantee of long term compatibility/support.
+	pamlessAuthStack := config.Section("Unstable").Key("pamless_auth_stack").MustBool(false)
```
ChaitanyaKulkarni28, line 103: do we also want to add section for this in `instance_configs.cfg` with default/empty value?

google_guest_agent/oslogin.go
```diff
@@ -276,6 +285,31 @@ func writeNSSwitchConfig(enable bool) error {
 	return writeConfigFile("/etc/nsswitch.conf", proposed)
 }
 
+func updatePAMsshdPamless(pamsshd string, enable, twofactor bool) string {
```
ChaitanyaKulkarni28, line 288: can we add unit test for this?

google_guest_agent/oslogin.go
dorileo: The default value is set to disabled, so it makes sense only to have config entries on testing deployments that will either happen manually or by deriving images, so I don't think we need.

google_guest_agent/oslogin.go
dorileo: Sure, will do.

google_guest_agent/main.go
```diff
@@ -98,6 +98,7 @@ func closeFile(c io.Closer) {
 func runUpdate(ctx context.Context) {
 	var wg sync.WaitGroup
 	mgrs := []manager{&addressMgr{}}
+
```
ericdand, line 101: Suggest reverting the change to this file, since you only add a newline.

google_guest_agent/oslogin.go
```diff
@@ -99,15 +99,18 @@ func (o *osloginMgr) set(ctx context.Context) error {
 		logger.Infof("Disabling OS Login")
 	}
 
-	if err := writeSSHConfig(enable, twofactor, skey); err != nil {
+	// Unstable configuration section is a section with no guarantee of long term compatibility/support.
```
ericdand, line 102: What does this comment mean?

dorileo: We don't have now a mechanism for feature flags or even development feature flags. This change is introducing a new mechanism for that, I'm making it more formalized with this [PR](https://github.com/GoogleCloudPlatform/guest-agent/pull/288#) - see the `type Unstable struct{}` part of it.

dorileo: I expanded the comments a bit further to make it clear. Since it's the first user of it I'll keep this comments in the same region of the code, I'll get it removed when we get the `cfg` package PR merged.

google_guest_agent/main.go
dorileo: Done. Thanks.

google_guest_agent/oslogin.go
dorileo: Done. Thanks.

google_guest_agent/oslogin.go
```diff
@@ -99,15 +99,20 @@ func (o *osloginMgr) set(ctx context.Context) error {
 		logger.Infof("Disabling OS Login")
 	}
 
-	if err := writeSSHConfig(enable, twofactor, skey); err != nil {
+	// [Unstable] configuration section has no o long term stability or support guarantees/promisses.
```
ericdand, line 102: Nice, this is much clearer. Thank you.

A couple of typo fixes on this line; it should read:

    [Unstable] configuration section has no long term stability or support guarantees/promises.

ericdand (approved):

PR #287, opened 2023-09-19
Preparing it for review

ChaitanyaKulkarni28:
Preparing it for review: Update repo to follow open source guidelines

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2020 Google Inc.
+   Copyright [yyyy] [name of copyright owner]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2020 Google Inc.
+   Copyright [yyyy] [name of copyright owner]
```
dorileo, line 189: I suppose we want to keep Google's copyright, don't we?

dorileo: BTW, since we are touching it we could update it to "Copyright 2023 Google LLC".

LICENSE
```diff
@@ -198,4 +198,4 @@
    distributed under the License is distributed on an "AS IS" BASIS,
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
-   limitations under the License.
+   limitations under the License.
```
dorileo, line 201: I guess we are just tweaking `/n` here, right? We could leave it as is.

PR #288, opened 2023-09-19
cfg: introduce a configuration management package

dorileo:


google_guest_agent/cfg/cfg_test.go
```diff
@@ -0,0 +1,92 @@
...
+		return []interface{}{
+			[]byte(invalidConfig),
+		}
+	}
+
+	if err := Load(); err == nil {
+		t.Errorf("Load() didn't fail to load invalid configuration, expected error")
+	}
+
+	dataSources = defaultDataSources
```
ChaitanyaKulkarni28, line 44: you can reset it with defer where its overridden on line 34 `defer dataSources = defaultDataSources`

google_guest_agent/instance_setup_integ_test.go
```diff
@@ -20,40 +20,53 @@ package main
...
 )
 
-// TestInstanceSetupSSHKeys validates SSH keys are generated on first boot and not changed afterward.
-func TestInstanceSetupSSHKeys(t *testing.T) {
-	cfg, err := parseConfig("") // get empty config
+func getConfig(t *testing.T) (*cfg.Sections, string) {
+	t.Helper()
+
+	if err := cfg.Load(nil); err != nil {
+		t.Fatal("Failed to load configuration: %+v", err)
```
ChaitanyaKulkarni28, line 38: `t.Fatalf`

google_guest_agent/instance_setup_integ_test.go
```diff
@@ -20,40 +20,53 @@ package main
...
+	}
+
+	config, err := cfg.Get()
 	if err != nil {
-		t.Fatal("failed to init config object")
+		t.Fatal("Failed to get config: %+v", err)
```
ChaitanyaKulkarni28, line 43: `t.Fatalf`

google_guest_agent/main.go
```diff
@@ -79,45 +76,41 @@ func logStatus(name string, disabled bool) {
...
+		mgrs = append(mgrs, []manager{newWsfcManager(config), &winAccountsMgr{}, &diagnosticsMgr{}}...)
 	default:
 		mgrs = append(mgrs, []manager{&clockskewMgr{}, &osloginMgr{}, &accountsMgr{}}...)
 	}
 	for _, mgr := range mgrs {
 		wg.Add(1)
 		go func(mgr manager) {
 			defer wg.Done()
-			if mgr.disabled(runtime.GOOS) {
+			if disabled, _ := mgr.Disabled(ctx, config); disabled {
```
ChaitanyaKulkarni28, line 99: I understand error is not used right now and all mgrs are returning nil, but it would be good to add error check right now so we don't have to remember about it when implementation actually starts throwing errors.

google_guest_agent/cfg/cfg.go
```diff
@@ -0,0 +1,299 @@
...
+	if err := cfg.MapTo(sections); err != nil {
+		return fmt.Errorf("failed to map configuration to object: %+v", err)
+	}
+
+	instance = sections
+	return nil
+}
+
+// Get returns the configuration's instance previously loaded with Load().
+func Get() (*Sections, error) {
```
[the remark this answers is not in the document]
dorileo: I wanted to make it very distinctive to what each call is doing, in guest-agent ideally we should have only one call to `Load()` and handled that as early init code. If we can't load defaults + user files we should abort - as a non recoverable error.

google_guest_agent/addresses_test.go
```diff
@@ -15,14 +15,32 @@
...
 
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/cfg"
 	"github.com/GoogleCloudPlatform/guest-agent/metadata"
-	"github.com/go-ini/ini"
 )
 
+func getConfig(t *testing.T, extraDefaults []byte) *cfg.Sections {
+	t.Helper()
+	if err := cfg.Load(extraDefaults); err != nil {
+		t.Errorf("Error parsing config: %+v", err)
```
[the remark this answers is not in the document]
dorileo: done, thanks.

google_guest_agent/instance_setup_integ_test.go
dorileo: done. thanks.

google_guest_agent/main.go
dorileo: done. thanks.

google_guest_agent/instance_setup_integ_test.go
dorileo: done. thanks.

google_guest_agent/cfg/cfg_test.go
dorileo: Done. thanks.

google_guest_agent/cfg/cfg.go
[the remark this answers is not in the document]
ChaitanyaKulkarni28: It can follow singleton pattern, load & return on first call and then reuse. That way there are no concerns about ensuring its loaded as early as possible.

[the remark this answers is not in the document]
dorileo: My point is handling the error condition for the two operations differently (different meaning) and in different stages of to the code execution.

google_guest_agent/cfg/cfg.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	instance = sections
+	return nil
+}
+
+// Get returns the configuration's instance previously loaded with Load().
+func Get() *Sections {
+	if instance == nil {
+		panic(fmt.Errorf("cfg package was not initialized, Load() " +
```
ChaitanyaKulkarni28, line 306: as we are not using `recover()` and this would result into string representation only, we could use direct string message instead of `fmt.Errorf`

google_guest_agent/cfg/cfg.go
```diff
@@ -0,0 +1,310 @@
...
+	return append(res, []interface{}{
+		[]byte(defaultConfig),
+		configFile,
+		configFile + ".distro",
+		configFile + ",template",
+	}...)
+}
+
+// Load loads default configuration and the configuration from default config files.
+func Load(extraDefaults []byte) error {
```
ChaitanyaKulkarni28, line 282: I am thinking instead of relying on the caller to `os.Exit()` or `panic` would it make sense to panic in `Load()` as well on error?

google_guest_agent/cfg/cfg.go
dorileo: Yep, we can. I'll get it updated. Thanks.

google_guest_agent/cfg/cfg.go
dorileo: This is one call that we can nicely handle - since there's supposed to be a single call in the whole guest-agent code base.

google_guest_agent/cfg/cfg.go
dorileo: Done. Thanks.

PR #289, opened 2023-09-20
sshca: re-enable the event watcher & handler

dorileo:
Additionally this change also fixes the AuthorizedPrincipalCommand configuration.


google_guest_agent/main.go
```diff
@@ -199,6 +202,8 @@ func runAgent(ctx context.Context) {
 		return
 	}
 
+	sshca.Init(eventManager)
```
ChaitanyaKulkarni28, line 205: do we need unstable config flag here or its ready for GA?

dorileo: We could protect it if we so decide to, however the MDS endpoint we are using is long available in prod and only having the pipe doesn't change the user journeys as we are not changing sshd config.

PR #290, opened 2023-09-20
Events interface improvement

dorileo:


google_guest_agent/events/events.go
```diff
@@ -84,224 +150,336 @@ type WatcherEventType struct {
...
+	for _, curr := range tmpSlice {
+		res = append(res, curr)
+	}
+
+	return res
 }
 
 // Run runs the event manager, it will block until all watchers have given up/failed.
-func (mngr *Manager) Run(ctx context.Context) {
+func (mngr *Manager) Run(ctx context.Context) error {
```
ChaitanyaKulkarni28, line 379: on return shouldn't this be closing all channels (like new one's introduced in `queue` structure) its listening on?

google_guest_agent/events/events.go
```diff
@@ -48,14 +49,65 @@ type Watcher interface {
...
+	// finishCallbackHandler is a channel used to communicate with the callback handling
+	// go routine that it should finish/end its job (usually after all watchers are done).
+	finishCallbackHandler chan bool
+
+	// watcherDone is a channel used to communicate that a given watcher is finished/done.
+	watcherDone chan string
+
+	// dataBus is the channel used to communicate between watchers (event producer) and the
+	// callback handler (event consumer managing go routine).
+	dataBus chan eventBusData
```
ChaitanyaKulkarni28, line 106: We might not be super sensitive about latency and all, but being a channel of length 1 wouldn't all other watchers wait to write new response until previous one is processed? would this be a bottleneck?

google_guest_agent/events/events.go
```diff
@@ -48,14 +49,65 @@ type Watcher interface {
 // Manager defines the interface between events management layer and the
 // core guest agent implementation.
 type Manager struct {
```
ChaitanyaKulkarni28, line 51: can we use `sync.RWMutex` here for any of the attributes? In some cases parallel reads should be fine, we can protect just the writes?

google_guest_agent/events/events.go
```diff
@@ -84,224 +150,336 @@ type WatcherEventType struct {
...
+
+	mngr.subscribersMutex.Lock()
+	defer mngr.subscribersMutex.Unlock()
+
+	tmpSlice, found := mngr.subscribers[evType]
+	if !found {
+		return nil
+	}
+
+	for _, curr := range tmpSlice {
```
ChaitanyaKulkarni28, line 371: why are we iterating over `tmpSlice` and generating `res` again instead of directly returning `tmpSlice`?

google_guest_agent/events/events.go
```diff
@@ -84,224 +150,336 @@ type WatcherEventType struct {
...
+// context (derived from the one provided in the AddWatcher() call) and will have it canceled
+// after calling this method.
+func (mngr *Manager) RemoveWatcher(ctx context.Context, watcher Watcher) error {
+	mngr.watchersMutex.Lock()
+	defer mngr.watchersMutex.Unlock()
+
+	id := watcher.ID()
+	logger.Debugf("Got a request to remove watcher: %s", id)
+	if _, found := mngr.watchersMap[id]; !found {
+		return fmt.Errorf("Unknown Watcher(%s)", id)
```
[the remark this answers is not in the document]
dorileo: done everywhere. thanks.

google_guest_agent/events/events.go
```diff
@@ -84,224 +150,336 @@ type WatcherEventType struct {
...
-	var leaving bool
 
-	syncBus := make(chan eventBusData)
-	defer close(syncBus)
-
-	cancelContext := make(chan bool)
-	cancelCallback := make(chan bool)
+	mngr.runningMutex.Lock()
+	if mngr.running {
+		return fmt.Errorf("Tried calling event manager's Run() twice")
```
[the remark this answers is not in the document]
dorileo: Fixed. Thanks.

google_guest_agent/events/events.go
```diff
@@ -84,224 +150,336 @@ type WatcherEventType struct {
...
-			// Report back if we don't know the provided watcher id.
-			if _, found := availableWatchers[curr]; !found {
-				return nil, fmt.Errorf("invalid/unknown watcher id: %s", curr)
-			}
-			res.watchers[curr] = availableWatchers[curr]
-		}
+// Get allocates a new manager if one doesn't exists or returns the one previously allocated.
+func Get() *Manager {
+	if instance == nil {
+		instance = newManager()
```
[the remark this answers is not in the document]
dorileo: True, done. Thanks.

google_guest_agent/events/events.go
dorileo: For simplicity this initial implementation is intended to have the callback handling/calling queued up so we don't have a go routine for each but a single go routine dequeueing the event processing. We can sure optimize it later if we have a need/requirement or if it becomes a bottleneck, in that case it would make sense to have a go routine per callback in the queue etc. Do you think we need this optimization now?

google_guest_agent/events/events.go
dorileo: Yep, we don't need it at all, I was trying to be extremely cautions but new slice heads will be created when modifying the map so we don't even need this function at all. Thanks for bringing that up.

google_guest_agent/events/events.go
dorileo: Changed wherever it made sense. Thank you.

google_guest_agent/events/events.go
```diff
@@ -84,224 +150,327 @@ type WatcherEventType struct {
...
-	var leaving bool
-
-	syncBus := make(chan eventBusData)
-	defer close(syncBus)
 
-	cancelContext := make(chan bool)
-	cancelCallback := make(chan bool)
+	mngr.runningMutex.Lock()
+	if mngr.running {
+		return fmt.Errorf("tried calling event manager's Run() twice")
```
ChaitanyaKulkarni28, line 375: Unlock on `runningMutex` seems to be missing in this case

dorileo: true. Will get it fixed.

dorileo: done. thanks.

PR #291, opened 2023-09-20
Fix permissions on file created by SaferWriteFile()

ChaitanyaKulkarni28:
Temporary files are created with permission `600`

### DIFF
--- a/utils/main.go
+++ b/utils/main.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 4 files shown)


utils/main.go
```diff
@@ -187,6 +187,10 @@ func SaferWriteFile(content []byte, outputFile string) error {
 		return fmt.Errorf("unable to create temporary file under %q: %w", dir, err)
 	}
 
+	if err := os.Chmod(tmp.Name(), 0644); err != nil {
```
dorileo, line 190: can we make it even more generic and add a `mode` argument to `SaferWriteFile()`?

ChaitanyaKulkarni28: done

PR #295, opened 2023-09-21
managers: we accidentally disabled addressMgr, bring it back

dorileo:


google_guest_agent/main.go
```diff
@@ -89,12 +89,14 @@ func closeFile(c io.Closer) {
 func availableManagers() []manager {
 	if runtime.GOOS == "windows" {
 		return []manager{
+			&addressMgr{},
```
ChaitanyaKulkarni28, line 92: since we have a common manager, do you want to keep a list of common managers and append to it based on os type?

dorileo: Could be, I can change it.

dorileo: done. thanks.

google_guest_agent/main.go
```diff
@@ -87,18 +87,25 @@ func closeFile(c io.Closer) {
...
 			newWsfcManager(),
 			&winAccountsMgr{},
 			&diagnosticsMgr{},
-		}
+		)
 	}
-	return []manager{
+
+	return append(managers,
+		&addressMgr{},
```
ChaitanyaKulkarni28, line 104: looks like we're appending `addressMgr` here as well. Same in case of windows

dorileo: indeed. fixed. thanks.

PR #296, opened 2023-09-22
Add support for the SSH key options

oleksiyivanenko:
SSH Key options as defined by the spec:
https://man.openbsd.org/sshd.8#agent-forwarding

For instance they allow authorized SSH Key to look something like:

```
restrict,command="echo hi" ssh-rsa [KEY_MATERIAL] [COMMENT]
```

To enable this:
- Switch function that parses the key from parsing by space to usage of ssh.ParseAuthorizedKey
- Add more tests that cover the new use case
- Fix all existing tests to have a valid rsa key material

### DIFF
--- a/google_authorized_keys/main_test.go
+++ b/google_authorized_keys/main_test.go
@@ -23,6 +23,7 @@ import (
 	"testing"
 
 	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-agent/utils"
 )
 
 func stringSliceEqual(a, b []string) bool {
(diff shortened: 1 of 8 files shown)


utils/main_test.go
```diff
@@ -15,11 +15,30 @@
...
+	"encoding/base64"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
+
+	"golang.org/x/crypto/ssh"
 )
 
+func mkRandRSAPubKey(t *testing.T) string {
```
dorileo, line 29: Can you please move this function to a non OS specific go file and avoid the duplication?

google_authorized_keys/main_test.go
```diff
@@ -166,6 +196,11 @@ func TestGetUserKeysNew(t *testing.T) {
 	}
 
 	for count, tt := range tests {
+		replaceAttributesKeys(t, &tt.instanceMetadata, "[KEY]", randKey)
```
dorileo, line 199: Can you please take the opportunity that you are already touching theses tests and wrap these set tests in a t.Run()? You can use the slice index to name the test i.e. fmt.Sprintf("test-%d", índex)

utils/main_test.go
```diff
@@ -37,76 +56,91 @@ func TestContainsString(t *testing.T) {
...
-		{`usera:ssh-rsa AAAA1234 google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`,
-			"usera", `ssh-rsa AAAA1234 google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`, false},
+		{`usera:ssh-rsa [KEY] google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`,
+			"usera", `ssh-rsa [KEY] google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`, false},
+		{`usera:restrict,pty ssh-rsa [KEY] google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`,
+			"usera", `restrict,pty ssh-rsa [KEY] google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`, false},
 		{"    ", "", "", true},
-		{"ssh-rsa AAAA1234", "", "", true},
-		{":ssh-rsa AAAA1234", "", "", true},
+		{"ssh-rsa [KEY]", "", "", true},
```
dorileo, line 72: Can you initialize the test data with the key rather than transforming it close to where you are using it? It will make the test easier to read (please apply everywhere you are following this pattern).

utils/main_test.go
```diff
@@ -37,76 +56,91 @@ func TestContainsString(t *testing.T) {
...
 		key    string
 		haserr bool
 	}{
-		{"usera", `ssh-rsa AAAA1234 google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`, false},
-		{"user a", `ssh-rsa AAAA1234 google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`, true},
-		{"usera", `ssh-rsa AAAA1234 google-ssh {"userName":"usera@example.com","expireOn":"2021-04-23T12:34:56+0000"}`, true},
-		{"usera", `ssh-rsa AAAA1234 google-ssh {"userName":"usera@example.com","expireOn":"Apri 4, 2056"}`, true},
-		{"usera", `ssh-rsa AAAA1234 google-ssh`, true},
-		{"usera", `ssh-rsa AAAA1234 test info`, false},
+		{"usera", `ssh-rsa [KEY] google-ssh {"userName":"usera@example.com","expireOn":"2095-04-23T12:34:56+0000"}`, false},
```
dorileo, line 97: Can you please group these test cases by the prefix up to the key token?

dorileo (changes requested):

utils/main_test.go
oleksiyivanenko: Created utils/test.go file to host this test specific util function. Please let me know if you'd prefer another place for this function

google_authorized_keys/main_test.go
oleksiyivanenko: Done

utils/main_test.go
oleksiyivanenko: Done

utils/main_test.go
oleksiyivanenko: Done

dorileo:
It's looking good, I'll hold it until we make a release this week and then have another pass. thanks.

PR #300, opened 2023-09-26
Refactor authorized keys to use metadata package

ChaitanyaKulkarni28:


google_authorized_keys/main.go
```diff
@@ -20,23 +20,24 @@ import (
...
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
 
-var (
-	programName       = "GoogleAuthorizedKeysCommand"
-	metadataURL       = "http://169.254.169.254/computeMetadata/v1/"
-	metadataRecursive = "/?recursive=true&alt=json"
-	defaultTimeout    = 2 * time.Second
-)
+const programName = "GoogleAuthorizedKeysCommand"
```
dorileo, line 34: I don't know why we were originally using `GoogleAuthorizedKeysCommand` here as it should be the binary name i.e. `path.Base(os.Args[0])`.

ChaitanyaKulkarni28: Done

google_authorized_keys/main.go
```diff
@@ -49,43 +52,14 @@ func logFormatWindows(e logger.LogEntry) string {
 	return fmt.Sprintf("%s %s: %s", now, programName, e.Message)
 }
 
-func getMetadata(key string, recurse bool) ([]byte, error) {
-	client := &http.Client{
-		Timeout: defaultTimeout,
-	}
+func getMetadata(ctx context.Context, key string) ([]byte, error) {
```
dorileo, line 55: Is it worth keeping this wrapper?

dorileo (commented):
Just minor comment, looking good.

google_authorized_keys/main.go
ChaitanyaKulkarni28: after refactoring it wasn't really doing much. Got rid of it. Thanks!

PR #302, opened 2023-10-03
Disable the IO scheduler based on metadata.

ilc-fg:
Because the google_guest_agent runs after udev rules, it overrides them, preventing the user from having any way to override the defaults, the way they might wish.

This disables that behavior allowing the "normal" way of setting the io scheduler to work, udev rules, etc.  With no concerns that GCP will come back and reset the scheduler.


ilc-fg:
Closing the PR.  It'll come back in another form.

PR #314, opened 2023-10-12
Update known cert dirs for updaters

ChaitanyaKulkarni28:


google_guest_agent/agentcrypto/mtls_mds_linux.go
```diff
@@ -51,39 +69,41 @@ func (j *CredsJob) writeRootCACert(ctx context.Context, content []byte, outputFi
 
 // writeClientCredentials stores client credentials (certificate and private key).
 func (j *CredsJob) writeClientCredentials(plaintext []byte, outputFile string) error {
+	if err := os.MkdirAll(filepath.Dir(outputFile), 0655); err != nil {
```
dorileo, line 72: the requirement is the same, so for consistency sake add the same comment here: "// The directory should be executable, but the file does not need to be."

dorileo (commented):
PR looking good, just minor picky comment.

google_guest_agent/agentcrypto/mtls_mds_linux.go
ChaitanyaKulkarni28: done

PR #315, opened 2023-10-16
Update response struct, add tests

ChaitanyaKulkarni28:
* Accommodate the changes in response structure for `instance/gce-workload-certificates/workload-identities` and `instance/gce-workload-certificates/trust-anchors` endpoint
* Add unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,15 @@
-# ignore all built binaries
+# Ignore all built binaries.
 **/gce_workload_cert_refresh
+**/gce_workload_cert_refresh.exe
 **/google_authorized_keys
+**/google_authorized_keys.exe
 **/google_guest_agent
+**/google_guest_agent.exe
 **/google_metadata_script_runner
+**/google_metadata_script_runner.exe
+
+# Don't ignore new content to directories.
+!**/gce_workload_cert_refresh/
+!**/google_authorized_keys/
+!**/google_guest_agent/
+!**/google_metadata_script_runner/
--- a/gce_workload_cert_refresh/main.go
+++ b/gce_workload_cert_refresh/main.go
@@ -23,6 +23,8 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
+	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 # ignore all built binaries
-**/gce_workload_cert_refresh
-**/google_authorized_keys
-**/google_guest_agent
-**/google_metadata_script_runner
+*/gce_workload_cert_refresh
```
dorileo, line 2: Why?

dorileo: I guess the right fix would be:

```patch
+ !**/gce_workload_cert_refresh/
+ !*/google_authorized_keys/
+ !*/google_guest_agent/
+ !*/google_metadata_script_runner/
+ **/*.exe
```

ChaitanyaKulkarni28: Done

gce_workload_cert_refresh/main.go
```diff
@@ -79,125 +89,62 @@ func getMetadata(ctx context.Context, key string) ([]byte, error) {
...
+		},
+        "PEER_SPIFFE_TRUST_DOMAIN_2": {
+            "trustAnchorsPem" : "<Trust bundle containing the X.509 roots certificates>",
+        }
+    }
+}
+*/
 
-	return nil
+// TrustAnchor represents a Root Certificate in metadata
```
tavishvaidya, line 134: The TrustAnchorsPem can contain non-root, CA certificates as well. In this case, the certificate chain validation will stop at that certificate. Also, the TrustAnchorsPem can contain multiple certificates one after the another in no particular order. Recommend updating the comment to remove Root and clarifying that TrustAnchor can contain multiple certificates.

gce_workload_cert_refresh/main_test.go
tavishvaidya, line 1: Thanks for adding these tests, they were very much needed!

tavishvaidya (approved):

gce_workload_cert_refresh/main.go
ChaitanyaKulkarni28: Done

PR #318, opened 2023-10-19
snapshot: add context cancellation handling

dorileo:
If snapshot is enabled and the process gets a SIG ABORT the context will be canceled but the snapshot listener will keep trying to create a context nevertheless. This change makes sure we return in case CreateConnection()'s returned error is a context cancellation.

### DIFF
--- a/google_guest_agent/snapshot_listener.go
+++ b/google_guest_agent/snapshot_listener.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


google_guest_agent/snapshot_listener.go
```diff
@@ -82,6 +83,7 @@ func listenForSnapshotRequests(ctx context.Context, address string, requestChan
 		r, err := c.CreateConnection(ctx, &guestReady)
 		if err != nil {
 			logger.Errorf("Error creating connection: %v.", err)
+			leaving = errors.Is(err, context.Canceled)
```
ChaitanyaKulkarni28, line 86: do we need to check `context.DeadlineExceeded` error as well? I mean stop this loop if its  `context.DeadlineExceeded` or `context,Canceled`?

PR #322, opened 2023-10-25
Remove previous MDS root certificate if refreshed

ChaitanyaKulkarni28:


google_guest_agent/agentcrypto/mtls_mds_windows.go
```diff
@@ -59,6 +59,11 @@ var (
 
 // writeRootCACert writes Root CA cert from UEFI variable to output file.
 func (j *CredsJob) writeRootCACert(_ context.Context, cacert []byte, outputFile string) error {
+	num, err := serialNumber(outputFile)
```
dorileo, line 62: Can we move this closer to line 92? It only makes sense there, the logging can also be done within that if statement.

dorileo:
Just nit picking comment, otherwise looking good.

google_guest_agent/agentcrypto/mtls_mds_windows.go
ChaitanyaKulkarni28: if we receive new creds, this file will be overwritten by that time. So we need to make sure previous one is read before we attempt to write

dorileo: Can you add a comment explaining that?

ChaitanyaKulkarni28: Done

PR #323, opened 2023-10-26
Enable mTLS cred generation by default

ChaitanyaKulkarni28:
Adds a section for MDS mTLS configuration and enables cred generation by default

### DIFF
--- a/google_guest_agent/cfg/cfg.go
+++ b/google_guest_agent/cfg/cfg.go
@@ -87,6 +87,9 @@ setup = true
 [OSLogin]
 cert_authentication = true
 
+[MDS]
+mtls_bootstrapping_enabled = true
+
 [Snapshots]
 enabled = false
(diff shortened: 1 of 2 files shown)


google_guest_agent/cfg/cfg.go
```diff
@@ -87,14 +87,16 @@ setup = true
 [OSLogin]
 cert_authentication = true
 
+[MDSmTLS]
```
dorileo, line 90: How about:

```
[MDS]
mtls_bootstrapping_enabled =  true
```

or maybe:
```
[Metadata]
mtls_bootstrapping_enabled =  true
```

ChaitanyaKulkarni28: Done, updated to use - 
```
[MDS]
mtls_bootstrapping_enabled =  true
```

PR #324, opened 2023-10-26
Test configuration files are loaded in the documented order. Fix initial integration test.

a-crate:
I said this was going to be a unit test but it's testing the locations from defaultDataSources so it is actually an integration test. In addition the fact that user configs are overridden by distro and template configs I also found that `extraDefault` values are overridden by `defaultConfigs`. The intention for these variables is not documented so I made an assumption based on the names, let me know if any of these test cases are wrong.

/assign @dorileo


google_guest_agent/cfg/cfg_integ_test.go
```diff
@@ -0,0 +1,101 @@
+//  Copyright 2023 Google LLC
+//
```
dorileo, line 2: This is not exactly integration test, adding those to cfg_test.go should be just fine.

google_guest_agent/cfg/cfg_integ_test.go
```diff
@@ -0,0 +1,101 @@
...
+			extraDefault:   "",
+			distroConfig:   "",
+			templateConfig: "",
+			userConfig:     "",
+			output:         true,
+		},
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := os.WriteFile(configFile+".distro", []byte(tc.distroConfig), 0777)
```
dorileo, line 80: Please use testing.T.TempDir() to guarantee these files are being removed, also to avoid messing the environment please make defaultConfigFile() swappable like defaultDataSource() (and make defaultDataSource() to use the pointer) so we can swap it for the test case and make it return the right file name constructed based on the tests temp dir.

dorileo (changes requested):

google_guest_agent/cfg/cfg_test.go
```diff
@@ -92,3 +96,81 @@ func TestGetTwice(t *testing.T) {
 		t.Errorf("Get() should return always the same pointer, expected: %p, got: %p", firstCfg, secondCfg)
 	}
 }
+
+func TestConfigLoadOrder(t *testing.T) {
+	config := path.Join(t.TempDir(), "config.cfg")
+	configFile = func(string) string { return config }
```
dorileo, line 102: Set it back to default after using it.

dorileo (changes requested):

google_guest_agent/cfg/cfg_test.go
a-crate: Oops, thought I did so. Fixed.

PR #326, opened 2023-10-31
Implement json-based command messaging system for guest-agent

a-crate:


google_guest_agent/main.go
```diff
@@ -175,6 +176,9 @@ func runAgent(ctx context.Context) {
 
 	agentInit(ctx)
 
+	cmdsrv := command.Init(ctx)
```
dorileo, line 179: we want to protect this initialization with a feature flag, it'd look like this:

```
if cfg.Get().Unstable.CommandMonitorEnabled {
  cmdServer := command.Init(ctx)
  defer cmdServer.Close()
}
```

google_guest_agent/command/command.go
```diff
@@ -0,0 +1,140 @@
+//  Copyright 2023 Google Inc. All Rights Reserved.
```
dorileo, line 1: what's the difference between `command.go` and `command_monitor.go`?

google_guest_agent/command/command.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	// HandlerError is returned when the handler function returns an non-nil error. The status message will be replaced with the returnd error string.
+	HandlerError = Response{
+		Status:        105,
+		StatusMessage: "The command handler encountered an error processing your request",
+	}
+	// InternalErrorCode is the error code for internal command server errors. Returned when failing to marshal a response.
+	InternalErrorCode = 106
+	internalError     = []byte(`{"Status":106,"StatusMessage":"The command server encountered an internal error trying to respond to your request"}`)
+	handlersMu        = new(sync.RWMutex)
```
dorileo, line 80: Can we wrap these `handlersMu` and `handlers` in a instance object?

google_guest_agent/command/command_monitor.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+// Init starts an internally managed command server which will begin listening
+// when handlers register, and stop listening when all handlers unregister. The
+// agent configuration will decide the server options. Returns a reference to
+// the internally managed command server which the caller can Close() when
+// appropriate.
+func Init(ctx context.Context) *Server {
+	pipe := cfg.Get().Unstable.CommandPipePath
```
dorileo, line 45: You are reinitializing the singleton instance every time we call `Init()`. You can do right off the begining the following:

```
if cmdServer != nil {
   return cmdServer
}
```

That's your signal to wether `Init()` was called before or not.

google_guest_agent/command/command_monitor.go
```diff
@@ -0,0 +1,263 @@
...
+		logger.Errorf("could not parse command_pipe_mode as octal integer: %v falling back to mode 0770", err)
+	}
+	cmdserver = NewCmdServer(pipe, int(pipemode), cfg.Get().Unstable.CommandPipeGroup, to)
+	go func() {
+		err := cmdserver.Wait(ctx)
+		if err != nil {
+			logger.Errorf("stopped waiting for commands: %v", err)
+		}
+	}()
+	handlersMu.RLock()
```
dorileo, line 66: You don't need to protect the `cmdServer.Start()` call as it will only be called once and before the callers acquire the server's pointer.

google_guest_agent/command/command_monitor.go
```diff
@@ -0,0 +1,263 @@
...
+	if len(handlers) > 0 {
+		cmdserver.Start()
+	}
+	return cmdserver
+}
+
+// NewCmdServer returns a pointer to a new Server listening on pipe p. Few
+// callers should be using this outside of unit testing, Init should be called
+// once and Server should be managed internally instead.
+func NewCmdServer(p string, fm int, group string, to time.Duration) *Server {
```
dorileo, line 77: you don't need/want to export `NewCmdServer()` it's an internal implementation detail and used only internally in the `command` package.

google_guest_agent/command/command_monitor.go
```diff
@@ -0,0 +1,263 @@
...
+		pipeGroup: group,
+		timeout:   to,
+		srvMu:     new(sync.Mutex),
+		signal:    make(chan string, 1),
+	}
+	return &cs
+}
+
+var (
+	cmdserver *Server
```
dorileo, line 90: Can you please move this to the top of the program? can you also rename it to `cmdServer` ?

google_guest_agent/command/command_monitor.go
```diff
@@ -0,0 +1,263 @@
...
+// on the given context, as long as it is valid. To signal the server to start
+// or stop listening, call Server.Start() or Server.Stop(). To signal the
+// Server to stop waiting, call Server.Close(). This function is
+// thread-safe, but calling it multiple times is generally a bad idea, as Close()
+// will only close a single Wait() call.
+func (c *Server) Wait(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
dorileo, line 221: we are gracefully returning, just log it and return nil. The caller is logging the error so only return an error when it's actually an error.

google_guest_agent/command/command.go
```diff
@@ -0,0 +1,140 @@
...
+// intialized and there are no more handlers registered, the server will be
+// signalled to stop listening for commands.
+func UnregisterHandler(cmd string) error {
+	if _, ok := handlers[cmd]; !ok {
+		return fmt.Errorf("cmd %s is not registered", cmd)
+	}
+	handlersMu.Lock()
+	defer handlersMu.Unlock()
+	delete(handlers, cmd)
+	if len(handlers) == 0 && cmdserver != nil {
```
dorileo, line 109: I'm not sure we want to stop the server in this condition, we don't have the use case now, but this kind of interface expects that the program may want to register handlers at any time, otherwise a registering code would have to check if the server is running or not. The bottomline is: the main goroutine initializes the server and leaves it functional for any component to register handlers at any point, it will only shut it down when the program is aborted/killed etc.

google_guest_agent/command/command_monitor.go
```diff
@@ -0,0 +1,263 @@
...
+				c.srvMu.Lock()
+				if c.srv == nil {
+					logger.Debugf("starting command server")
+					err := c.listen(ctx)
+					if err != nil {
+						logger.Errorf("could not listen for commands on pipe %s: %v", c.pipe, err)
+					}
+				}
+				c.srvMu.Unlock()
+			case "STOP":
```
dorileo, line 234: Given my previous comment about Stop, would you consider we still need START and STOP signals? or even CLOSE?

google_guest_agent/command/command_monitor.go
```diff
@@ -0,0 +1,263 @@
...
+					c.srv = nil
+				}
+				c.srvMu.Unlock()
+				return nil
+			}
+		}
+	}
+}
+
+func marshalOrInternalError(v any) []byte {
```
dorileo, line 256: Why not make it `func marshal(v any) ([]byte, error)` ? I know the calling will be more verbose but the interface will be cleaner and easier to follow.

dorileo (changes requested):

google_guest_agent/main.go
a-crate: My intention was to write something that didn't require an explicit enablement flag, as if some subsystem X depends on command monitor functionality this would allow it to start listening whenever X was enabled (and registered its handlers with the monitor) and otherwise do nothing. This was the same reason for the Start/Stop signals. We don't have a flag to disable the events system either, I was considering this as a similar system, but we can move to a feature flag gated one.

google_guest_agent/command/command.go
a-crate: `command.go` contains the API for handlers (registration, deregistration, CmdNotFound status codes, etc). They are all the same package of course, I just split it in such a way that someone reading through the API to implement a handler wouldn't be seeing implementation details.

google_guest_agent/command/command.go
a-crate: I don't see a way to implement this cleanly in the Register and Unregister calls, since cmdServer is nil if handlers register before it's initialized and we don't want callers besides the main agent calling `Init()`
```
// With a feature flag
if !cfg.Get().Unstable.CommandMonitorEnabled {
  return fmt.Errorf("disabled")
}
for cmdServer == nil {
}
cmdServer.handlers[cmd] = handler
```
But the infinite loop until the server exists strikes me as a bad pattern.

google_guest_agent/command/command_monitor.go
a-crate: Done.

google_guest_agent/command/command_monitor.go
a-crate: Done.

google_guest_agent/command/command_monitor.go
a-crate: Done.

google_guest_agent/command/command_monitor.go
a-crate: Done.

google_guest_agent/command/command.go
a-crate: >otherwise a registering code would have to check if the server is running or not

Not really, if we just take out `Start` and `Stop`, replace them with a `CommandServerEnabled` toggle, and listen for the entire server lifetime will still work as expected and allowing registration and deregistration at any time. The commandServer will pick up new handlers on each new connection.

google_guest_agent/command/command_monitor.go
a-crate: We will need Close to gracefully `Close()` `c.srv` anytime the agent shuts down to unlink the unix socket. But yeah Start and Stop were to create the possibility to decide automatically when to listen, they are not necessary otherwise.

google_guest_agent/command/command_monitor.go
a-crate: There's no reason for it to exist with a signature like that, it might as well just be `json.Marshal` then. The only use is an internal helper function to do `conn.Write(marshalOrInternalError(v))`. If you prefer calling `json.Marshal`, I can switch.

google_guest_agent/command/command.go
dorileo: Ok, that should be fine.

google_guest_agent/command/command.go
dorileo: We are exposing `Server` right? We should protect `Server` as server is an implementation detail and expose the command monitor. So you can have:

```
type Monitor struct {
   server Server
   handlersMu sync.RWMutex
   handlers map[string]Handler
}
```
Note: you could un-export server though since we want to hide it.

then you would have a:

```
func Init(...) Monitor {
    return &Monitor{server: initServer(), ....}
}
```
Where `initServer()` would be your current `Init()` implementation.

Or either, to be consistent with other packages we don't return `Monitor` but initialize the singleton variable and use a `Get()` to return the instance.

With that the calling pattern would look like this:

```
if err := command.Init(); err != nil {
    log.Error(err)
}
defer command.Close()
....

command.Get().Subscribe()
...
```

`command.Close()` would use the singleton variable.

PS: if we have a singleton `Monitor` we don't need the singleton `Server`.

google_guest_agent/command/command_monitor.go
dorileo: Yep, definitely better avoiding the wrapper in this case.

google_guest_agent/main.go
dorileo: I quite don't follow this `as if some subsystem X depends on command monitor functionality this would allow it to start listening whenever X was enabled`, what do you mean? Having a feature flag allows customers to disable it entirely if they need/want, if we can't allow it then we are doing something wrong and should fix. PS: Initially we want to keep it disabled by default until we launch the feature.

dorileo (changes requested):

google_guest_agent/main.go
a-crate: I mean, imagine this configuration for some future subsystem of guest-agent that depends on command-monitor:

```
[X]
enabled = true
[CommandMonitor]
enabled = false
```

If X needs the command monitor for core functionality, how should it handle it? Just log an error? Disable itself? Should subsystems never design around guaranteed availability of the command monitor? We would need to document this, or at least document that subsystems should direct users to enable command monitor if it's required.

The intention for my current design is that if X is the only user of command monitor, then it will open the socket when `X.Enabled` is true and will not open it when `X.Enabled` is false.

google_guest_agent/command/command.go
a-crate: ACK. I misunderstood what you were saying at first, I see now.

google_guest_agent/main.go
dorileo: If subsystem X exposes a command and command monitor is disabled then the `Register()` call will be a no-op - yes, we can log that it resulted in a no-op.

Subsystem X can expose its own configuration section and expose a `enabled` flag, from a user perspective it's a "inclusive or" where Subsystem X and command monitor enabled flags must be true to enable it.

In short, a command monitor enabled flag set to false will disable all commands.

a-crate: Ack. I'll include some documentation for this.

google_guest_agent/main.go
```diff
@@ -175,6 +176,11 @@ func runAgent(ctx context.Context) {
 
 	agentInit(ctx)
 
+	if cfg.Get().Unstable.CommandMonitorEnabled {
+		cmdsrv := command.Init(ctx)
```
dorileo, line 180: One last minor bit, change `command.Init()` to return nothing (maybe error only if that's the case) and add a `command.Close()`, the `command.Init()` would set the `cmdMonitor` pointer then `command.Close()` will check if `cmdMonitor` pointer is nil, if it's nil then do nothing, if not nil then we do call  `cmdMonitor.Close()` and set the pointer to nil. For other parts of the guest-agent using it we can just use the `command.Get()` function (that's already returning `cmdMonitor`.

dorileo (approved):

PR #331, opened 2023-11-08
Update script runner to use common cfg package

ChaitanyaKulkarni28:
Metadata script runner should use common cfg package for parsing/fetching `instance_configs.cfg` values rather than introducing its own logic.

### DIFF
--- a/google_metadata_script_runner/main.go
+++ b/google_metadata_script_runner/main.go
@@ -38,15 +38,13 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/cfg"
 	"github.com/GoogleCloudPlatform/guest-agent/metadata"
 	"github.com/GoogleCloudPlatform/guest-agent/utils"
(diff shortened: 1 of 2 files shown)


google_metadata_script_runner/main.go
```diff
@@ -428,11 +433,13 @@ func main() {
...
 	var err error
-	config, err = parseConfig(cfgfile)
-	if err != nil && !os.IsNotExist(err) {
-		fmt.Printf("Error parsing instance config %s: %s\n", cfgfile, err.Error())
+	if err := cfg.Load(nil); err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to load instance configuration: %+v", err)
+		os.Exit(1)
 	}
 
+	config = cfg.Get()
```
dorileo, line 441: can we stop using this `config` variable and simply use `cfg.Get()` in the 3 or 4 callers?

dorileo (changes requested):

google_metadata_script_runner/main.go
ChaitanyaKulkarni28: Done

dorileo (approved):

PR #336, opened 2023-12-14
Fix snapshot test failure

dorileo:


google_guest_agent/snapshot_listener.go
```diff
@@ -77,21 +77,23 @@ func listenForSnapshotRequests(ctx context.Context, address string, requestChan
 
 		c := sspb.NewSnapshotServiceClient(conn)
 		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
```
a-crate, line 80: If you defer the cancellation of this child context to the exit of the function but only break the loop (and exit the function) when the parent context is cancelled the lifetime of the child context is identical to the lifetime of the parent context. Is this intentional? If so why do you even need the child context?

dorileo: Yep, we don't. Just adjusted. Thanks.

a-crate (approved):

PR #337, opened 2023-12-18
Add github.com/Microsoft/go-winio to third party licensing

a-crate:
https://github.com/microsoft/go-winio/blob/main/LICENSE
/assign @dorileo


dorileo:
side note: we should automate checking for new deps added in presubmit.

PR #341, opened 2024-01-06
vendor dependencies

dorileo:
As a requirement coming from Mozilla licensing we should also ship/carry the source code for dependencies, we have a 2nd level dependency that forces us to do so for that purpose we are adding an explicit dependency (github.com/hashicorp/hcl) so we can use standard tooling without having to deal with exception in the tooling level.


dorileo:
we have a couple of things with this PR:
1) if we are vendoring stuff we shouldn't create a exception process, we should always use the standard tooling - in this case go mod vendor;
2) if we want to vendor a 2+1 level dependency then we must implicitly import it in the code so go mod tidy + go mod vendor remains functional - the exception is dealt in this level see the change in `google_guest_agent/agentcrypto/crypto_util.go` in this PR;

dorileo:
If we decide to go with this we need to adjust golint to skip vendor dir.

PR #342, opened 2024-01-11
Deprecate DHClient in favor of systemd-networkd

drewhli:
This is part of the ongoing work to deprecate DHClient. The main priority currently is utilizing `systemd-networkd`, since Ubuntu will be removing `dhclient` after their upcoming freeze at the end of January. The other network management services, such as `wicked` and `NetworkManager` will come later.

Currently, this will do the following
- If `dhclient` is found on the system, we continue using `dhclient` for network interface configuration.
- If not, then we use `systemd-networkd`.

### DIFF
--- a/google_guest_agent/addresses.go
+++ b/google_guest_agent/addresses.go
@@ -19,15 +19,13 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"os"
 	"reflect"
 	"runtime"
 	"strings"
-	"time"
 
 	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/cfg"
+	network "github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/network/manager"
 	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
-	"github.com/GoogleCloudPlatform/guest-agent/metadata"
 	"github.com/GoogleCloudPlatform/guest-agent/utils"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
@@ -102,20 +100,6 @@ func compareRoutes(configuredRoutes, desiredRoutes []string) (toAdd, toRm []stri
 }
 
 var badMAC []string
-
-func getInterfaceByMAC(mac string) (net.Interface, error) {
-	hwaddr, err := net.ParseMAC(mac)
-	if err != nil {
-		return net.Interface{}, err
-	}
-
-	for _, iface := range interfaces {
-		if iface.HardwareAddr.String() == hwaddr.String() {
-			return iface, nil
-		}
-	}
-	return net.Interface{}, fmt.Errorf("no interface found with MAC %s", mac)
-}
 
 // https://www.ietf.org/rfc/rfc1354.txt
 // Only fields that we currently care about.
@@ -296,22 +280,10 @@ func (a *addressMgr) Set(ctx context.Context) error {
 		return fmt.Errorf("error populating interfaces: %v", err)
 	}
 
-	if config.NetworkInterfaces.Setup {
-		if runtime.GOOS != "windows" {
-			logger.Debugf("Configure IPv6")
-			if err := configureIPv6(ctx); err != nil {
-				// Continue through IPv6 configuration errors.
-				logger.Errorf("Error configuring IPv6: %v", err)
-			}
-		}
-
-		if runtime.GOOS != "windows" && !interfacesEnabled {
-			logger.Debugf("Enable network interfaces")
-			if err := enableNetworkInterfaces(ctx, config); err != nil {
-				return err
-			}
-			interfacesEnabled = true
-		}
+	// Setup network interfaces.
+	err = network.SetupInterfaces(ctx, config, newMetadata.Instance.NetworkInterfaces)
+	if err != nil {
+		return fmt.Errorf("failed to setup network interfaces: %v", err)
 	}
 
 	if !config.NetworkInterfaces.IPForwarding {
@@ -321,7 +293,7 @@ func (a *addressMgr) Set(ctx context.Context) error {
 	logger.Debugf("Add routes for aliases, forwarded IP and target-instance IPs")
 	// Add routes for IP aliases, forwarded and target-instance IPs.
 	for _, ni := range newMetadata.Instance.NetworkInterfaces {
-		iface, err := getInterfaceByMAC(ni.Mac)
+		iface, err := network.GetInterfaceByMAC(ni.Mac)
 		if err != nil {
 			if !utils.ContainsString(ni.Mac, badMAC) {
 				logger.Errorf("Error getting interface: %s", err)
@@ -444,194 +416,4 @@ func (a *addressMgr) Set(ctx context.Context) error {
 	}
 
 	return nil
-}
-
-// Enables or disables IPv6 on network interfaces.
-func configureIPv6(ctx context.Context) error {
-	var newNi, oldNi metadata.NetworkInterfaces
-	if len(newMetadata.Instance.NetworkInterfaces) == 0 {
-		return fmt.Errorf("no interfaces found in metadata")
-	}
-	newNi = newMetadata.Instance.NetworkInterfaces[0]
-	if len(oldMetadata.Instance.NetworkInterfaces) > 0 {
-		oldNi = oldMetadata.Instance.NetworkInterfaces[0]
-	}
-	iface, err := getInterfaceByMAC(newNi.Mac)
-	if err != nil {
-		return err
-	}
-	switch {
-	case oldNi.DHCPv6Refresh != "" && newNi.DHCPv6Refresh == "",
-		newNi.DHCPv6Refresh == "" && len(oldMetadata.Instance.NetworkInterfaces) == 0:
-		// disable
-		// uses empty old interface slice to indicate this is first-run.
-
-		// Before obtaining or releasing an IPv6 lease, we wait for
-		// 'tentative' IPs as part of SLAAC. We wait up to 5 seconds
-		// for this condition to automatically resolve.
-		tentative := []string{"-6", "-o", "a", "s", "dev", iface.Name, "scope", "link", "tentative"}
-		for i := 0; i < 5; i++ {
-			res := run.WithOutput(ctx, "ip", tentative...)
-			if res.ExitCode == 0 && res.StdOut == "" {
-				break
-			}
-			time.Sleep(1 * time.Second)
-		}
-		if err := run.Quiet(ctx, "dhclient", "-r", "-6", "-1", "-v", iface.Name); err != nil {
-			return err
-		}
-	case oldNi.DHCPv6Refresh == "" && newNi.DHCPv6Refresh != "":
-		// enable
-		tentative := []string{"-6", "-o", "a", "s", "dev", iface.Name, "scope", "link", "tentative"}
-		for i := 0; i < 5; i++ {
-			res := run.WithOutput(ctx, "ip", tentative...)
-			if res.ExitCode == 0 && res.StdOut == "" {
-				break
-			}
-			time.Sleep(1 * time.Second)
-		}
-		val := fmt.Sprintf("net.ipv6.conf.%s.accept_ra_rt_info_max_plen=128", iface.Name)
-		if err := run.Quiet(ctx, "sysctl", val); err != nil {
-			return err
-		}
-		if err := run.Quiet(ctx, "dhclient", "-1", "-6", "-v", iface.Name); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-// enableNetworkInterfaces runs `dhclient eth1 eth2 ... ethN`
-// and `dhclient -6 eth1 eth2 ... ethN`.
-// On RHEL7, it also calls disableNM for each interface.
-// On SLES, it calls enableSLESInterfaces instead of dhclient.
-func enableNetworkInterfaces(ctx context.Context, config *cfg.Sections) error {
-	if len(newMetadata.Instance.NetworkInterfaces) < 2 {
-		return nil
-	}
-	var googleInterfaces []string
-	// The primary (first) interface is managed by the OS, we only handle
-	// secondary interfaces in this code.
-	for _, ni := range newMetadata.Instance.NetworkInterfaces[1:] {
-		iface, err := getInterfaceByMAC(ni.Mac)
-		if err != nil {
-			if !utils.ContainsString(ni.Mac, badMAC) {
-				logger.Errorf("Error getting interface: %s", err)
-				badMAC = append(badMAC, ni.Mac)
-			}
-			continue
-		}
-		googleInterfaces = append(googleInterfaces, iface.Name)
-	}
-	var googleIpv6Interfaces []string
-	for _, ni := range newMetadata.Instance.NetworkInterfaces[1:] {
-		if ni.DHCPv6Refresh == "" {
-			// This interface is not IPv6 enabled
-			continue
-		}
-		iface, err := getInterfaceByMAC(ni.Mac)
-		if err != nil {
-			if !utils.ContainsString(ni.Mac, badMAC) {
-				logger.Errorf("Error getting interface: %s", err)
-				badMAC = append(badMAC, ni.Mac)
-			}
-			continue
-		}
-		googleIpv6Interfaces = append(googleIpv6Interfaces, iface.Name)
-	}
-
-	switch {
-	case osInfo.OS == "sles":
-		return enableSLESInterfaces(ctx, googleInterfaces)
-	case (osInfo.OS == "rhel" || osInfo.OS == "centos") && osInfo.Version.Major >= 7:
-		for _, iface := range googleInterfaces {
-			err := disableNM(iface)
-			if err != nil {
-				return err
-			}
-		}
-		fallthrough
-	default:
-		dhcpCommand := config.NetworkInterfaces.DHCPCommand
-		if dhcpCommand != "" {
-			tokens := strings.Split(dhcpCommand, " ")
-			return run.Quiet(ctx, tokens[0], tokens[1:]...)
-		}
-
-		// Try IPv4 first as it's higher priority.
-		if err := run.Quiet(ctx, "dhclient", googleInterfaces...); err != nil {
-			return err
-		}
-
-		if len(googleIpv6Interfaces) == 0 {
-			return nil
-		}
-		for _, iface := range googleIpv6Interfaces {
-			// Enable kernel to accept to route advertisements.
-			val := fmt.Sprintf("net.ipv6.conf.%s.accept_ra_rt_info_max_plen=128", iface)
-			if err := run.Quiet(ctx, "sysctl", val); err != nil {
-				return err
-			}
-		}
-
-		var dhclientArgs6 []string
-		dhclientArgs6 = append([]string{"-6"}, googleIpv6Interfaces...)
-		return run.Quiet(ctx, "dhclient", dhclientArgs6...)
-	}
-}
-
-// enableSLESInterfaces writes one ifcfg file for each interface, then
-// runs `wicked ifup eth1 eth2 ... ethN`
-func enableSLESInterfaces(ctx context.Context, interfaces []string) error {
-	var err error
-	var priority = 10100
-	for _, iface := range interfaces {
-		logger.Debugf("write enabling ifcfg-%s config", iface)
-
-		var ifcfg *os.File
-		ifcfg, err = os.Create("/etc/sysconfig/network/ifcfg-" + iface)
-		if err != nil {
-			return err
-		}
-		defer closer(ifcfg)
-		contents := []string{
-			googleComment,
-			"STARTMODE=hotplug",
-			// NOTE: 'dhcp' is the dhcp4+dhcp6 option.
-			"BOOTPROTO=dhcp",
-			fmt.Sprintf("DHCLIENT_ROUTE_PRIORITY=%d", priority),
-		}
-		_, err = ifcfg.WriteString(strings.Join(contents, "\n"))
-		if err != nil {
-			return err
-		}
-		priority += 100
-	}
-	args := append([]string{"ifup", "--timeout", "1"}, interfaces...)
-	return run.Quiet(ctx, "/usr/sbin/wicked", args...)
-}
-
-// disableNM writes an ifcfg file with DHCP and NetworkManager disabled.
-func disableNM(iface string) error {
-	logger.Debugf("write disabling ifcfg-%s config", iface)
-	filename := "/etc/sysconfig/network-scripts/ifcfg-" + iface
-	ifcfg, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0644)
-	if err == nil {
-		defer closer(ifcfg)
-		contents := []string{
-			googleComment,
-			fmt.Sprintf("DEVICE=%s", iface),
-			"BOOTPROTO=none",
-			"DEFROUTE=no",
-			"IPV6INIT=no",
-			"NM_CONTROLLED=no",
-			"NOZEROCONF=yes",
-		}
-		_, err = ifcfg.WriteString(strings.Join(contents, "\n"))
-		return err
-	}
-	if os.IsExist(err) {
-		return nil
-	}
-	return err
 }
--- a/google_guest_agent/network/manager/common.go
+++ b/google_guest_agent/network/manager/common.go
@@ -1,0 +1,86 @@
+//  Copyright 2024 Google LLC
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package manager is responsible for detecting the current network manager service, and
+// writing and rolling back appropriate configurations for each network manager service.
+package manager
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+var (
+	badMAC = make(map[string]net.Interface)
+)
+
+// interfaceNames extracts the names of the network interfaces from the provided list
+// of network interfaces.
+func interfaceNames(nics []metadata.NetworkInterfaces) ([]string, error) {
+	var ifaces []string
+	for _, ni := range nics {
+		iface, err := GetInterfaceByMAC(ni.Mac)
+		if err != nil {
+			return nil, err
+		}
+		ifaces = append(ifaces, iface.Name)
+	}
+	return ifaces, nil
+}
+
+// interfaceListsIpv4Ipv6 gets a list of interface names. The first list is a list of all
+// interfaces, and the second list consists of only interfaces that support IPv6.
+func interfaceListsIpv4Ipv6(nics []metadata.NetworkInterfaces) ([]string, []string) {
+	var googleInterfaces []string
+	var googleIpv6Interfaces []string
+
+	for _, ni := range nics {
+		iface, err := GetInterfaceByMAC(ni.Mac)
+		if err != nil {
+			if _, found := badMAC[ni.Mac]; !found {
+				logger.Errorf("error getting interface: %s", err)
+				badMAC[ni.Mac] = iface
+			}
+			continue
+		}
+		if ni.DHCPv6Refresh != "" {
+			googleIpv6Interfaces = append(googleIpv6Interfaces, iface.Name)
+		}
+		googleInterfaces = append(googleInterfaces, iface.Name)
+	}
+	return googleInterfaces, googleIpv6Interfaces
+}
+
+// GetInterfaceByMAC gets the interface given the mac string.
+func GetInterfaceByMAC(mac string) (net.Interface, error) {
+	hwaddr, err := net.ParseMAC(mac)
+	if err != nil {
+		return net.Interface{}, err
+	}
+
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return net.Interface{}, fmt.Errorf("failed to get interfaces: %v", err)
+	}
+
+	for _, iface := range interfaces {
+		if iface.HardwareAddr.String() == hwaddr.String() {
+			return iface, nil
+		}
+	}
+	return net.Interface{}, fmt.Errorf("no interface found with MAC %s", mac)
+}
--- a/google_guest_agent/network/manager/dhclient_linux.go
+++ b/google_guest_agent/network/manager/dhclient_linux.go
@@ -1,0 +1,285 @@
+//  Copyright 2024 Google LLC
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package manager is responsible for detecting the current network manager service, and
+// writing and rolling back appropriate configurations for each network manager service.
+package manager
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/cfg"
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/ps"
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+const (
+	// The base directory for dhclient files managed by guest agent.
+	baseDhclientDir = "/var/google-dhclient.d"
+
+	// For finer control of the execution, dhclient is invoked for
+	// each interface individually such that each call will have its
+	// own PID file. This is where those PID files are expected to be
+	// written.
+	pidFileDir = "pids"
+
+	// Similar thing for PID files, but for DHClient leases.
+	leaseFileDir = "leases"
+)
+
+var (
+	// ipv4 is a wrapper containing the protocol version and its respective
+	// dhclient argument.
+	ipv4 = ipVersion{"ipv4", "-4"}
+
+	// ipv6 is a wrapper containing the protocol version and its respective
+	// dhclient argument.
+	ipv6 = ipVersion{"ipv6", "-6"}
+)
+
+// ipVersion is a wrapper containing the human-readable version string and
+// the respective dhclient argument.
+type ipVersion struct {
+	// version is the human-readable IP protocol version.
+	version string
+
+	// dhclientArg is the respective argument for DHClient invocation.
+	dhclientArg string
+}
+
+// dhclient implements the manager.Service interface for dhclient use cases.
+type dhclient struct{}
+
+// init adds this network manager service to the list of known network managers.
+// DHClient will be the default fallback.
+func init() {
+	registerManager(&dhclient{}, true)
+}
+
+// Name returns the name of the network manager service.
+func (n dhclient) Name() string {
+	return "dhclient"
+}
+
+// IsManaging checks if the dhclient CLI is available.
+func (n dhclient) IsManaging(ctx context.Context, iface string) (bool, error) {
+	// Check if the dhclient CLI exists.
+	if _, err := exec.LookPath("dhclient"); err != nil {
+		if errors.Is(err, exec.ErrNotFound) {
+			return false, nil
+		}
+		return false, fmt.Errorf("error looking up dhclient path: %v", err)
+	}
+	return true, nil
+}
+
+// Setup sets up the non-primary interfaces with dhclient, having different setup procedures
+// for IPv6 network interfaces and IPv4 network interfaces.
+func (n dhclient) Setup(ctx context.Context, config *cfg.Sections, payload []metadata.NetworkInterfaces) error {
+	dhcpCommand := config.NetworkInterfaces.DHCPCommand
+	if dhcpCommand != "" {
+		tokens := strings.Split(dhcpCommand, " ")
+		return run.Quiet(ctx, tokens[0], tokens[1:]...)
+	}
+
+	// Create the necessary directories.
+	if err := os.MkdirAll(fmt.Sprintf("%s/%s", baseDhclientDir, pidFileDir), 0755); err != nil {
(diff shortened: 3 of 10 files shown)


google_guest_agent/network/dhclient.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-agent/utils"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+type dhclient struct{}
+
+func (n dhclient) IsManaging(ctx context.Context, iface string) (bool, error) {
+	// Check if the dhclient CLI exists.
+	result := run.WithOutput(ctx, "command", "-v", "dhclient")
```
ChaitanyaKulkarni28, line 18: You can use `exec.LookPath`  to verify existence of that command.

google_guest_agent/network/dhclient.go
```diff
@@ -0,0 +1,109 @@
...
+	if result.ExitCode == 1 {
+		// Failed to find dhclient command.
+		return false, nil
+	} else if result.ExitCode != 0 {
+		// Something went wrong when executing the command.
+		return false, fmt.Errorf("error getting dhclient command: %v", result.StdErr)
+	}
+
+	// Check for the process pertaining the dhclient managing the primary interface.
+	result = run.WithOutput(ctx, "/bin/sh", "-c", "ps aux | grep dhclient")
```
ChaitanyaKulkarni28, line 28: Do you need to pass `/bin/sh -c`? I believe you should be able to run `ps` command directly

google_guest_agent/network/managers.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+var (
+	badMAC               = []string{}
+	googleComment        = "# Added by Google Compute Engine OS Login."
+	managedKey           = "ManagedByGuestAgent=true"
+	knownNetworkManagers = []ManagerService{dhclient{}, systemdnetworkd{}}
+)
+
+func commandExists(ctx context.Context, cmd string) (bool, error) {
```
ChaitanyaKulkarni28, line 27: Same as above: you can use exec.LookPath to verify existence of that command.

google_guest_agent/network/managers.go
```diff
@@ -0,0 +1,85 @@
...
+		iface, err := getInterfaceByMAC(ni.Mac)
+		if err != nil {
+			return nil, err
+		}
+		ifaces = append(ifaces, iface.Name)
+	}
+	return ifaces, nil
+}
+
+func getInterfaceByMAC(mac string) (net.Interface, error) {
```
ChaitanyaKulkarni28, line 52: I believe this method is very similar to `getInterfaceByMAC()` in `address.go` can you consolidate to avoid code duplication?

google_guest_agent/network/managers.go
```diff
@@ -0,0 +1,85 @@
...
+// using different network managing services, such as systemd-networkd and wicked.
+type ManagerService interface {
+	IsManaging(ctx context.Context, iface string) (bool, error)
+	Setup(ctx context.Context, payload []metadata.NetworkInterfaces) error
+	Rollback(ctx context.Context, payload []metadata.NetworkInterfaces) error
+}
+
+var (
+	badMAC               = []string{}
+	googleComment        = "# Added by Google Compute Engine OS Login."
```
ChaitanyaKulkarni28, line 22: Should this be `Guest Agent` instead of `OS Login`?

google_guest_agent/network/managers.go
```diff
@@ -0,0 +1,85 @@
...
+			return iface, nil
+		}
+	}
+
+	return net.Interface{}, fmt.Errorf("no interface found with MAC %s", mac)
+}
+
+// DetectNetworkManager Detects the network manager managing the primary network interface.
+// This network manager will be used to set up secondary network interfaces.
+func DetectNetworkManager(ctx context.Context, iface string) (ManagerService, error) {
```
ChaitanyaKulkarni28, line 74: can we cache this value instead of trying to detect every time?

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+const configDir = "/etc/systemd/network"
+
+type systemdnetworkd struct{}
+
+func (n systemdnetworkd) IsManaging(ctx context.Context, iface string) (bool, error) {
+	// First check if the service is running.
+	res := run.WithOutput(ctx, "/bin/sh", "-c", "systemctl status systemd-networkd.service | grep Active:")
```
ChaitanyaKulkarni28, line 20: same as above, i don't think you need to pass `/bin/sh` for every command

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (n systemdnetworkd) Rollback(ctx context.Context, payload []metadata.NetworkInterfaces) error {
+	// Delete the configuration files created by the agent.
+	// Read the directory.
+	f, err := os.Open(configDir)
+	if err != nil {
+		return fmt.Errorf("failed to open systemd-networkd configuration directory: %v", err)
+	}
+	files, err := f.Readdir(-1)
+	f.Close()
```
ChaitanyaKulkarni28, line 88: you're missing error check on this `close()`

google_guest_agent/network/dhclient.go
drewhli: This is so we can use the pipe to `grep` have the output be anything containing `dhclient`. This avoids having to loop through each line of the output and check for matches ourselves.

google_guest_agent/network/systemd_networkd.go
drewhli: Done.

google_guest_agent/network/managers.go
drewhli: Done.

google_guest_agent/network/managers.go
drewhli: Done.

google_guest_agent/network/dhclient.go
drewhli: Done.

google_guest_agent/network/managers.go
drewhli: Done.

google_guest_agent/network/managers.go
drewhli: I think we still have to detect it every time. But caching the current one will be useful for rolling back the changes of one manager service in order to set the changes for the new one, in cases where the primary network managing service gets changed.

google_guest_agent/addresses.go
```diff
@@ -552,31 +540,24 @@ func enableNetworkInterfaces(ctx context.Context, config *cfg.Sections) error {
...
-		if len(googleIpv6Interfaces) == 0 {
-			return nil
-		}
-		for _, iface := range googleIpv6Interfaces {
-			// Enable kernel to accept to route advertisements.
-			val := fmt.Sprintf("net.ipv6.conf.%s.accept_ra_rt_info_max_plen=128", iface)
-			if err := run.Quiet(ctx, "sysctl", val); err != nil {
-				return err
+		if manager != currManagerService {
+			if err = currManagerService.Rollback(ctx, newMetadata.Instance.NetworkInterfaces); err != nil {
```
dorileo, line 550: you only want to Rollback if the `currManager` is no `nil`.

drewhli: Done.

dorileo:
We are still missing the ipv6 configuration code path. It's still using `dhclient` see the function `configureIPv6()` in `addresses.go`.

google_guest_agent/network/manager.go
```diff
@@ -0,0 +1,39 @@
...
+
+// DetectNetworkManager Detects the network manager managing the primary network interface.
+// This network manager will be used to set up secondary network interfaces.
+func DetectNetworkManager(ctx context.Context, iface string) (ManagerService, error) {
+	for _, curr := range knownNetworkManagers {
+		ok, err := curr.IsManaging(ctx, iface)
+		if err != nil {
+			return nil, err
+		}
+		if ok {
```
dorileo, line 34: Also here, please add something line:

```
logger.Infof("Network manager detected(%+v): %#v", ok, curr)
```

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+const configDir = "/etc/systemd/network"
+
+type systemdnetworkd struct{}
+
+func (n systemdnetworkd) IsManaging(ctx context.Context, iface string) (bool, error) {
+	// First check if the service is running.
+	res := run.WithOutput(ctx, "/bin/sh", "-c", "systemctl status systemd-networkd.service | grep Active:")
```
dorileo, line 20: Not the most effective way of checking if a unit is active, please consider doing the following:

```
  res := run.WithOutput(ctx, "/bin/sh", "-c", "systemctl is-active systemd-networkd.service")
  if res.ExitCode != 0 {
    return false, nil
  }
```

If the unit is active it will return a 0 exit code, it will return exit code 1 otherwise.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+	if res.ExitCode != 0 {
+		return false, fmt.Errorf("failed to get status of systemd-networkd service: %v", res.StdErr)
+	}
+	if !strings.Contains(res.StdOut, "running") {
+		// Service is not running, so it can't be managing an interface.
+		return false, nil
+	}
+
+	// Check systemd network configuration.
+	res = run.WithOutput(ctx, "/bin/sh", "-c", "networkctl status | grep "+iface)
```
dorileo, line 30: To check if systemd-networkd is managing an interface please consider doing the following:

```
  res = run.WithOutput(ctx, "/bin/sh", "-c", fmt.Sprintf("networkctl list %s --json=short", iface))
  if res.ExitCode != 0 {
    return false, fmt.Errorf("failed to check systemd-networkd network status: %v", res.StdErr)
  }

  interfaceStatus := make(map[string]any)

  if err := json.Unmarshal([]byte(res.StdOut), &interfaceStatus); err != nil {
    return false, fmt.Errorf("failed to unmarshal interface status: %+v", err)
  }

  adminstrativeState, found := interfaceStatus["AdministrativeState"]
  if !found {
    return false, fmt.Errorf("could not determine interface state, AdministrativeState is not present")
  }

  return adminstrativeState == "configured", nil
```

That way we don't trust solely on the stdout format but in a machine readable format (json in this case). We also make sure that we are querying only the interface of interest (the primary one).

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+		// has set.
+		config, err = os.Create(fmt.Sprintf("%s/50-%s.network", configDir, iface))
+		if err != nil {
+			return err
+		}
+		defer config.Close()
+
+		// Default, simple configuration for any network interface.
+		contents := []string{
+			googleComment,
```
dorileo, line 66: What's the purpose of this `googleComment`? just inform the users that we are managing this file?

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	for _, iface := range ifaces {
+		logger.Debugf("write network config for %s", iface)
+
+		var config *os.File
+
+		// The number in beginning of the file name signifies its priority. We want this
+		// to take priority before any of the default configurations that systemd-networkd
+		// has set.
+		config, err = os.Create(fmt.Sprintf("%s/50-%s.network", configDir, iface))
```
dorileo, line 58: Please use a known suffix for this file name to create extra signal that we are managing this file ourselves and not assume only based on the interface name. Consider something like: `50-<interface name>-google-guest-agent.network`.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to open systemd-networkd configuration directory: %v", err)
+	}
+	files, err := f.Readdir(-1)
+	f.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close file reader: %v", err)
+	}
+
+	for _, file := range files {
```
dorileo, line 98: You don't need to iterate over files in the dir as we can deterministically try to open the expected file, we are managing the file name pattern (see my comment above about file naming).

google_guest_agent/network/manager.go
```diff
@@ -0,0 +1,39 @@
+package network
```
dorileo, line 1: we are missing the network package comment.

google_guest_agent/network/manager.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+)
+
+// ManagerService is an interface for setting up network configurations
+// using different network managing services, such as systemd-networkd and wicked.
+type ManagerService interface {
+	IsManaging(ctx context.Context, iface string) (bool, error)
```
dorileo, line 13: Please add a comment/doc to each of these methods.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+		return false, fmt.Errorf("failed to check systemd-networkd network status: %v", res.StdErr)
+	}
+	if strings.Contains(res.StdOut, "configured") {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (n systemdnetworkd) Name() string {
+	return "systemd-networkd"
```
dorileo, line 41: please add a field to `systemdnetworkd` structure and use it instead of hardcoding inside the method.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+const configDir = "/etc/systemd/network"
+
+type systemdnetworkd struct{}
+
+func (n systemdnetworkd) IsManaging(ctx context.Context, iface string) (bool, error) {
```
dorileo, line 18: Please add comments/doc to this function and describe the strategy we are following.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+const configDir = "/etc/systemd/network"
```
dorileo, line 14: `configDir` is a very generic term and will certainly clash with other network managers implementation when they come up. Please also consider adding a field to `systemdnetworkd` structure rather than simply using a const.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+
+type systemdnetworkd struct{}
```
dorileo, line 16: please consider naming it `systemdNetworkd`.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+	if res.ExitCode != 0 {
+		return false, fmt.Errorf("failed to check systemd-networkd network status: %v", res.StdErr)
+	}
+	if strings.Contains(res.StdOut, "configured") {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (n systemdnetworkd) Name() string {
```
dorileo, line 40: Even that it's pretty obvious what this method is doing please consider adding comments/docs to it mentioning that it's the implementation of the interface etc.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+	return "systemd-networkd"
+}
+
+func (n systemdnetworkd) Setup(ctx context.Context, payload []metadata.NetworkInterfaces) error {
```
dorileo, line 44: Please add comments/docs describing what the `Setup()` operation means for `systemd-networkd` implementation.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	// Restart the systemd-networkd service.
+	logger.Debugf("restarting systemd-networkd")
+	return run.Quiet(ctx, "systemctl", "restart", "systemd-networkd.service")
+}
+
+func (n systemdnetworkd) Rollback(ctx context.Context, payload []metadata.NetworkInterfaces) error {
```
dorileo, line 85: Please add comments/docs describing what consists the `Rollback()` operation for the `systemd-netword` implementation.

google_guest_agent/network/utils.go
```diff
@@ -0,0 +1,41 @@
+package network
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+)
+
+// Gets a list of interface names.
```
dorileo, line 10: please follow the format of `// <function name> <description>`.

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+
+		// Default, simple configuration for any network interface.
+		contents := []string{
+			googleComment,
+			managedKey,
+			"[Match]",
+			fmt.Sprintf("Name=%s", iface),
+			"",
+			"[Network]",
+			"DHCP=yes",
```
dorileo, line 72: Are we checking somewhere if we should enable ipv6?

google_guest_agent/network/systemd_networkd.go
```diff
@@ -0,0 +1,112 @@
...
+			"IPForward=true",
```
dorileo, line 73: Why do we need this `IPForward` for? and do we need them to always be true?

google_guest_agent/network/systemd_networkd.go
drewhli: Yes.

google_guest_agent/network/systemd_networkd.go
drewhli: Done.

google_guest_agent/network/manager.go
drewhli: Done.

google_guest_agent/network/manager.go
drewhli: Done.

google_guest_agent/network/manager/dhclient.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// Rollback finds and kills the dhclient processes, effectively undoing the dhclient configurations.
+func (n dhclient) Rollback(ctx context.Context, payload []metadata.NetworkInterfaces) error {
+	// Find the dhclient processes.
+	ifaces, err := network.InterfaceNames(payload)
+	if err != nil {
+		return fmt.Errorf("failed to get interfaces: %s", err)
+	}
+	if len(ifaces) <= 1 { // nothing to roll back.
```
dorileo, line 121: this is not exactly correct (or true), right? we should still rollback/release the `dhclient -6` process.

drewhli: Yes, we should do the same for the primary interface.

drewhli: It's more along the lines of, this was an artifact of the previous implementation where we did not touch the primary interface at all in the Setup step.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,149 @@
...
+	googleInterfaces, googleIpv6Interfaces := InterfaceListsIpv4Ipv6(payload)
+
+	// Extra step for primary interface.
+	primaryIface := payload[0]
+	primaryNic, err := GetInterfaceByMAC(payload[0].Mac)
+	if err != nil {
+		return fmt.Errorf("error getting primary network interface: %s", err)
+	}
+
+	if err := ConfigureIPv6(ctx, googleIpv6Interfaces, primaryNic.Name); err != nil {
```
ChaitanyaKulkarni28, line 88: should this setup part on primary nic happen only if we know we won't be disabling it? (condition on line 93)

drewhli: The setup will not affect the primary interface unless its DHCPv6Refresh field is set. This is because it won't be an interface listed in `googleIpv6Interfaces`.

ChaitanyaKulkarni28: but `primaryNic` would still be set here right? So wouldn't `ConfigureIPv6` try to run `tentative := []string{"-6", "-o", "a", "s", "dev", primaryInterface, "scope", "link", "tentative"}`  on primary nic? Also, what if primary nic here does not support ipv6?

drewhli: Yes, it would still try to run it. I can see that we can skip the `tentative := ...` part if the primaryNic doesn't support support IPv6. If the primary nic doesn't support IPv6, currently it would just stop, which doesn't match the previously implemented behavior. Thanks for catching this.

drewhli:
Not exactly sure why the build is failing. The errors point to nonexistent functions that do, in fact, exist.

google_guest_agent/network/manager/systemd_networkd_linux.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			if os.IsNotExist(err) {
+				continue
+			} else {
+				return err
+			}
+		}
+
+		// Double check that the file has the guest-managed key before deleting it.
+		if strings.Contains(string(contents), guestAgentConfigIDKey) {
```
dorileo, line 148: this is a ini file, you can use `github.com/go-ini/ini` just like we are doing in `google_guest_agent/cfg/cfg.go`.

google_guest_agent/network/manager/systemd_networkd_linux.go
```diff
@@ -0,0 +1,156 @@
...
+		// to take priority before any of the default configurations that systemd-networkd
+		// has set.
+		config, err := os.Create(fmt.Sprintf("%s/50-%s-google-guest-agent.network", n.ConfigDir, iface))
+		if err != nil {
+			return err
+		}
+		defer config.Close()
+
+		// Default, simple configuration for any network interface.
+		contents := []string{
```
dorileo, line 107: this is a ini file, you can use `github.com/go-ini/ini` (see comment bellow for unmarshalling too).

dorileo (changes requested):

google_guest_agent/addresses.go
```diff
@@ -446,99 +424,13 @@ func (a *addressMgr) Set(ctx context.Context) error {
...
-}
-
-// enableNetworkInterfaces runs `dhclient eth1 eth2 ... ethN`
-// and `dhclient -6 eth1 eth2 ... ethN`.
-// On RHEL7, it also calls disableNM for each interface.
-// On SLES, it calls enableSLESInterfaces instead of dhclient.
+// enableNetworkInterfaces sets up network interfaces using the network manager that is
+// managing the primary network interface.
 func enableNetworkInterfaces(ctx context.Context, config *cfg.Sections) error {
 	if len(newMetadata.Instance.NetworkInterfaces) < 2 {
```
dorileo, line 430: I believe this check is preventing us to setup ipv6 for primary interface.

dorileo (changes requested):

google_guest_agent/addresses.go
drewhli: Done.

google_guest_agent/network/manager/systemd_networkd_linux.go
drewhli: Done.

google_guest_agent/network/manager/common.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+var (
+	badMAC = make(map[string]net.Interface)
+)
+
+// InterfaceListsIpv4Ipv6 gets a list of interface names. The first list is a list of all
+// non-primary interfaces, and the second list consists of only non-primary interfaces
```
ChaitanyaKulkarni28, line 32: Is this comment accurate about returning only secondary interfaces? Looks like code is iterating through all interfaces and not really skipping primary one

google_guest_agent/network/manager/common_linux.go
```diff
@@ -0,0 +1,72 @@
...
+		if err := run.Quiet(ctx, "sysctl", val); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// interfaceNames extracts the names of the network interfaces from the provided list
+// of network interfaces.
+func interfaceNames(nics []metadata.NetworkInterfaces) ([]string, error) {
```
ChaitanyaKulkarni28, line 62: since implementation is not specific to Linux should it be in `common.go` rather than `common_linux.go` ?

google_guest_agent/network/manager/systemd_networkd_linux.go
```diff
@@ -0,0 +1,169 @@
...
+		logger.Debugf("write %s network config for %s", n.Name(), iface)
+
+		var dhcp = "ipv4"
+		if slices.Contains(googleIpv6Interfaces, iface) {
+			dhcp = "yes"
+		}
+
+		// Create and setup ini file
+		config := ini.Empty()
+		guestSection, err := config.NewSection("GuestAgent")
```
ChaitanyaKulkarni28, line 97: nit: you can use variable `guestAgentConfigSection` defined in `manager.go`

google_guest_agent/network/manager/manager.go
```diff
@@ -0,0 +1,78 @@
...
+	// Setup writes the appropriate configurations for the network manager service for all
+	// non-primary network interfaces.
+	Setup(ctx context.Context, config *cfg.Sections, payload []metadata.NetworkInterfaces) error
+
+	// Rollback rolls back the changes created in Setup.
+	Rollback(ctx context.Context, payload []metadata.NetworkInterfaces) error
+}
+
+var (
+	googleComment = "# Added by Google Compute Engine Guest Agent."
```
ChaitanyaKulkarni28, line 46: nit: these should be constants instead of variables (except `knownNetworkManagers`)

google_guest_agent/network/manager/manager.go
```diff
@@ -0,0 +1,78 @@
...
+// using different network managing services, such as systemd-networkd and wicked.
+type Service interface {
+	// IsManaging checks whether this network manager service is managing the provided interface.
+	IsManaging(ctx context.Context, iface string) (bool, error)
+
+	// Name is the name of the network manager service.
+	Name() string
+
+	// Setup writes the appropriate configurations for the network manager service for all
+	// non-primary network interfaces.
```
ChaitanyaKulkarni28, line 38: nit: primary and secondary

google_guest_agent/network/manager/common.go
drewhli: Yea, I forgot to update the comment when I updated the method.

drewhli:
For context for the most recent commit, when testing on Ubuntu-2204, an error occurred where it could not find the `AdministrativeState` key. Upon manually running `networkctl status ens4 --json=pretty`, it seems that the key we need is different there compared to later versions of Ubuntu.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,191 @@
...
+	registerManager(dhclient{}, true)
+}
+
+// Name returns the name of the network manager service.
+func (n dhclient) Name() string {
+	return "dhclient"
+}
+
+// IsManaging checks if the dhclient CLI is available. If it is, then it proceeds to check
+// for the dhclient processes related to the provided interface. If the process  exists,
```
ChaitanyaKulkarni28, line 76: nit: can you also update the comment to avoid confusion. (we're no longer checking for specific interface)

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,191 @@
...
+
+	// Set up IPV4 for all secondary interfaces.
+	for _, iface := range googleInterfaces[1:] {
+		if err := runDhclient(ctx, ipv4, iface); err != nil {
+			logger.Errorf("failed to run dhclient: %+x", err)
+		}
+	}
+
+	// Setup IPV6 for all interfaces (primary included).
+	for _, iface := range googleIpv6Interfaces {
```
ChaitanyaKulkarni28, line 148: We're already calling `dhclient` on line 134 for primary interface, and since it is included here as well, wouldn't it run `dhclient` twice on it?

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,191 @@
...
+	return nil
+}
+
+// runDhclient releases and re-obtains a lease with the provided IP version
+// for the given network interface.
+func runDhclient(ctx context.Context, ipVersion ipVersion, nic string) error {
+	pidFile := path.Join(pidFileDir, fmt.Sprintf("dhclient.google-guest-agent.%s.pid", nic))
+	dhclientArgs := []string{ipVersion.dhclientArg, "-pf", pidFile, nic}
+
+	releaseArgs := append(dhclientArgs, "-r", nic)
```
ChaitanyaKulkarni28, line 162: I think we need to call `-r` with specific pid file as well, otherwise it would only try stop process with PID present in default location `/var/run/dhclient.pid`/`/var/run/dhclient6.pid`

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,191 @@
...
+func (n dhclient) Rollback(ctx context.Context, payload []metadata.NetworkInterfaces) error {
+	// Find the dhclient processes.
+	ifaces, err := interfaceNames(payload)
+	if err != nil {
+		return fmt.Errorf("failed to get interfaces: %s", err)
+	}
+
+	// Release all the interface leases from dhclient.
+	for _, iface := range ifaces {
+		if err = run.Quiet(ctx, "dhclient", "-r", iface); err != nil {
```
ChaitanyaKulkarni28, line 186: same as above: we need to specify pid file here

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+	}
+	return nil
+}
+
+// runDhclient releases and re-obtains a lease with the provided IP version
+// for the given network interface.
+func runDhclient(ctx context.Context, ipVersion ipVersion, nic string) error {
+	pidFile := path.Join(pidFileDir, fmt.Sprintf("dhclient.google-guest-agent.%s.pid", nic))
+	dhclientArgs := []string{ipVersion.dhclientArg, "-pf", pidFile, nic}
```
ChaitanyaKulkarni28, line 160: I wonder if we need to override lease file (`-lf`) as well? 

@dorileo

google_guest_agent/network/manager/dhclient_linux.go
drewhli: Done.

google_guest_agent/network/manager/dhclient_linux.go
drewhli: Yes. I've added a check to skip this now.

google_guest_agent/network/manager/dhclient_linux.go
drewhli: Done.

google_guest_agent/network/manager/dhclient_linux.go
drewhli: Done.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	for _, iface := range googleIpv6Interfaces {
+		// Set the appropriate system values.
+		val := fmt.Sprintf("net.ipv6.conf.%s.accept_ra_rt_info_max_plen=128", iface)
+		if err := run.Quiet(ctx, "sysctl", val); err != nil {
+			return err
+		}
+	}
+	switch {
```
ChaitanyaKulkarni28, line 124: we need to consolidate this logic as its not only for primary interfaces. Also, we should be calling `dhclient` the same way we call it for secondary interfaces (using `runDhclient`)

drewhli: Done.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,297 @@
...
+
+		// Check for IPv6 interfaces for which to obtain a lease.
+		// Do not setup IPv4 for primary interface.
+		if _, err := os.Stat(pidFilev4); i != 0 && os.IsNotExist(err) {
+			obtainIpv4Interfaces = append(obtainIpv4Interfaces, iface)
+		}
+
+		// Check for IPv6 interfaces for which to obtain a lease.
+		_, err := os.Stat(pidFilev6)
+		if os.IsNotExist(err) && slices.Contains(ipv6Interfaces, iface) {
```
ChaitanyaKulkarni28, line 269: Why are we checking if PID file exists and assigning IP only if doesn't? There could be a PID file but no process running. For e.g. on restart previous file would still be present but process is gone.

drewhli: Fixed to account for the extra case.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+		time.Sleep(1 * time.Second)
+	}
+
+	// Setup IPv6.
+	for _, iface := range obtainIpv6Interfaces {
+		// Set appropriate system values.
+		val := fmt.Sprintf("net.ipv6.conf.%s.accept_ra_rt_info_max_plen=128", iface)
+		if err := run.Quiet(ctx, "sysctl", val); err != nil {
+			return err
```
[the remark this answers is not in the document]
drewhli: The old execution still returned the error instead of logging. I think the only difference was that if it failed for the primary interface, then it'd continue with the secondary interfaces regardless, but if any of the secondary interfaces failed, then we'd stop and return. For consistent behavior across all the interfaces it's probably best to keep it as what we have here.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -0,0 +1,284 @@
...
+		// Only release if the flag is set.
+		releaseArgs := append(dhclientArgs, "-r", nic)
+		if err := run.Quiet(ctx, "dhclient", releaseArgs...); err != nil {
+			return fmt.Errorf("error releasing lease for %s: %v", nic, err)
+		}
+	} else {
+		// Now obtain a lease if release is not set.
+		dhclientArgs = append(dhclientArgs, nic)
+		if err := run.Quiet(ctx, "dhclient", dhclientArgs...); err != nil {
+			return err
```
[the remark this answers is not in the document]
drewhli: Done.

google_guest_agent/network/manager/systemd_networkd_linux.go
```diff
@@ -0,0 +1,236 @@
...
+		// The number in beginning of the file name signifies its priority. We want this
+		// to take priority before any of the default configurations that systemd-networkd
+		// has set.
+		configPath := fmt.Sprintf("%s/50-%s-google-guest-agent.network", n.configDir, iface)
+		if err := config.SaveTo(configPath); err != nil {
+			return fmt.Errorf("error saving config for %s: %v", iface, err)
+		}
+
+		// Enable the interface without restarting systemd-networkd service.
+		if err := run.Quiet(ctx, "networkctl", "up", iface); err != nil {
```
dorileo, line 192: Only calling networkctl up will reload it, I'd assume it requires a down and up sequence.

google_guest_agent/network/manager/systemd_networkd_linux.go
```diff
@@ -0,0 +1,240 @@
...
+
+		config := ini.Empty()
+		if err := ini.ReflectFrom(config, &data); err != nil {
+			return fmt.Errorf("error creating config ini: %v", err)
+		}
+
+		// The number in beginning of the file name signifies its priority. We want this
+		// to take priority before any of the default configurations that systemd-networkd
+		// has set.
+		configPath := fmt.Sprintf("%s/%s-%s-google-guest-agent.network", n.priority, n.configDir, iface)
```
ChaitanyaKulkarni28, line 191: shouldn't this be `n.configDir, n.priority, iface` as `n.configDir` is the directory name

google_guest_agent/network/manager/systemd_networkd_linux.go
```diff
@@ -0,0 +1,240 @@
...
+func (n systemdNetworkd) Rollback(ctx context.Context, payload []metadata.NetworkInterfaces) error {
+	logger.Infof("rolling back changes for %s", n.Name())
+	interfaces, err := interfaceNames(payload)
+	if err != nil {
+		return fmt.Errorf("failed to get list of interface names: %v", err)
+	}
+
+	for _, iface := range interfaces[1:] {
+		// Find expected files.
+		configFile := fmt.Sprintf("%s/%s-%s-google-guest-agent.network", n.priority, n.configDir, iface)
```
ChaitanyaKulkarni28, line 213: same as above

google_guest_agent/network/manager/systemd_networkd_linux.go
```diff
@@ -0,0 +1,240 @@
...
+				DHCP: dhcp,
+			},
+		}
+
+		config := ini.Empty()
+		if err := ini.ReflectFrom(config, &data); err != nil {
+			return fmt.Errorf("error creating config ini: %v", err)
+		}
+
+		// The number in beginning of the file name signifies its priority. We want this
```
ChaitanyaKulkarni28, line 188: nit: can you also mention lower the number == higher priority to avoid any future confusion

google_guest_agent/network/manager/manager.go
```diff
@@ -0,0 +1,360 @@
...
+	// agent should continue with the rest of the setup.
+	ignoreSetup bool
+}
+
+const (
+	googleComment = "# Added by Google Compute Engine Guest Agent."
+
+	// guestAgentConfigIDKey is added to the configurations written by Guest Agent
+	// to serve as an extra signal that the configuration is managed by the agent.
+	guestAgentConfigIDKey = "ManagedByGuestAgent"
```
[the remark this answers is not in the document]
drewhli: Done.

google_guest_agent/network/manager/manager.go
```diff
@@ -0,0 +1,360 @@
...
+	fallbackNetworkManager Service
+
+	// currManager is the Service implementation currently managing the interfaces.
+	currManager Service
+
+	// osRules lists the rules for applying interface configurations for primary
+	// and secondary interfaces.
+	osRules = []osConfigRule{
+		// Debian rules.
+		osConfigRule{
```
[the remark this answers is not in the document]
drewhli: Done.

google_guest_agent/network/manager/systemd_networkd_linux.go
drewhli: Yes. Done.

google_guest_agent/network/manager/systemd_networkd_linux.go
drewhli: Done.

google_guest_agent/network/manager/systemd_networkd_linux.go
drewhli: Done. Added more relevant details.

dorileo (approved):

PR #343, opened 2024-01-11
Delete integration tests

drewhli:
The integration tests have not been updated to reflect the changes to the guest agent that have been made since their previous update.

Despite the precheck saying that everything is good, for the past few months the `google_guest_agent` integration test builds have been failing. This aims to rectify that and show actual errors (...that only display when you click on the build details; it still doesn't actually reflect the failures here)

/hold


drewhli:
As of this comment, the `google_guest_agent` integration tests build, but are not passing.
https://oss.gprow.dev/view/gs/oss-prow/pr-logs/pull/GoogleCloudPlatform_guest-agent/343/guest-agent-integ/1745532115539202048

drewhli:
Decided to delete the integration tests for a few reasons
- They've been broken for a long time, despite the presubmits showing that they have been passing. Up until now they have been failing to build in the first place.
- We will be reimplementing these tests in CIT instead.

dorileo (approved):

PR #344, opened 2024-01-11
integration tests: these tests must exist in the CIT project

dorileo:
Rather than keeping some integration tests in guest-agent code base and others in cit we should keep it consistent and move it all over to cit.


dorileo:
closed in favor of https://github.com/GoogleCloudPlatform/guest-agent/pull/343

PR #348, opened 2024-02-02
Implement cleanup of deprecated configuration directives

dorileo:
Introduces a mechanism to cleanup configuration directives that we no longer support and that could be left behind after software update that could potentially harm the running system.

### DIFF
--- a/google_guest_agent/oslogin.go
+++ b/google_guest_agent/oslogin.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"slices"
 	"strings"
 	"time"
 
@@ -37,6 +38,13 @@ var (
 	googleBlockStart = "#### Google OS Login control. Do not edit this section. ####"
 	googleBlockEnd   = "#### End Google OS Login control section. ####"
 	trustedCAWatcher events.Watcher
+
+	// deprecatedConfigDirectives contains a list of configuration directives (or lines)
+	// that we no longer support and should not be considered for updated versions of a
+	// given configuration file.
+	deprecatedConfigDirectives = map[string][]string{
+		"/etc/pam.d/su": []string{"account    [success=bad ignore=ignore] pam_oslogin_login.so"},
+	}
 )
 
 type osloginMgr struct{}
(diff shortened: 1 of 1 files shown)


google_guest_agent/oslogin.go
```diff
@@ -188,6 +199,44 @@ func (o *osloginMgr) Set(ctx context.Context) error {
 	return nil
 }
 
+func cleanupDeprecatedLines(fpath string, directives []string) error {
+	// If the file doesn't exist don't even try updating it.
+	stat, err := os.Stat(fpath)
+	if os.IsNotExist(err) {
```
ChaitanyaKulkarni28, line 205: we still need to handle other errors right?

google_guest_agent/oslogin.go
```diff
@@ -121,6 +130,8 @@ func (o *osloginMgr) Set(ctx context.Context) error {
 	oldEnable, _, _ := getOSLoginEnabled(oldMetadata)
 	enable, twofactor, skey := getOSLoginEnabled(newMetadata)
 
+	cleanupDeprecatedDirectives()
```
ChaitanyaKulkarni28, line 133: should this be skipped if agent has updated once before (keeping a cache of file update)? Otherwise on every `Set()` agent will try finding diff and overwriting file with no change

google_guest_agent/oslogin.go
dorileo: Sure. I'll change it, thank you.

google_guest_agent/oslogin.go
dorileo: How about "only overwrite the content" if it changed, meaning, we couldn't find any deprecated content?

dorileo: done. thanks.

google_guest_agent/oslogin.go
dorileo: done. thanks.

google_guest_agent/oslogin.go
```diff
@@ -188,6 +199,55 @@ func (o *osloginMgr) Set(ctx context.Context) error {
...
+	// If the file doesn't exist don't even try updating it.
+	stat, err := os.Stat(fpath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to stat config file: %+v", err)
+	}
+
+	data, err := ioutil.ReadFile(fpath)
```
ChaitanyaKulkarni28, line 212: I just noticed we're using `ioutil`, I believe that's deprecated.

dorileo: good catch. thanks.

ChaitanyaKulkarni28, line 212: `ioutil` is deprecated, it seems you missed changing this while fixing line 232

dorileo: done. thanks.

PR #351, opened 2024-02-05
Update manager tests, Create DHClient unit tests.

drewhli:
- Create more comprehensive manager tests
- Add a unit test suite for dhclient.
- Refactor `run` and `ps` packages to provide publicly overrideable interfaces for ease and consistency in testing.
- Bugfix for `systemd-networkd` `Rollback()`: With including primary interfaces in the writing of the configuration files, those configuration files should also be included when rolling back.
- Bugfix for `dhclient` `dhclientProcessExists()`: When checking for the existence of an IPv6 dhclient process for a given interface, the current behavior would return true even if the process does not actually exist due to the fact that if we cannot locate the `-6` flag, then we return true by default provided we found the interface in the `dhclient` args list.

/hold

### DIFF
--- a/google_guest_agent/network/manager/common.go
+++ b/google_guest_agent/network/manager/common.go
@@ -19,6 +19,7 @@ package manager
 import (
 	"fmt"
 	"net"
+	"os/exec"
 
 	"github.com/GoogleCloudPlatform/guest-agent/metadata"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
@@ -26,6 +27,9 @@ import (
 
 var (
 	badMAC = make(map[string]net.Interface)
+
+	// execLookPath points to the function to check if a path exists.
+	execLookPath = exec.LookPath
 )
 
 // interfaceNames extracts the names of the network interfaces from the provided list
--- a/google_guest_agent/network/manager/dhclient_linux.go
+++ b/google_guest_agent/network/manager/dhclient_linux.go
@@ -36,7 +36,7 @@ import (
 
 const (
 	// The base directory for dhclient files managed by guest agent.
-	baseDhclientDir = "/var/google-dhclient.d"
+	defaultBaseDhclientDir = "/var/google-dhclient.d"
 
 	// For finer control of the execution, dhclient is invoked for
 	// each interface individually such that each call will have its
@@ -56,6 +56,11 @@ var (
 	// ipv6 is a wrapper containing the protocol version and its respective
 	// dhclient argument.
 	ipv6 = ipVersion{"ipv6", "-6"}
+)
+
+var (
+	// baseDhclientDir points to the base directory for DHClient leases and PIDs.
+	baseDhclientDir = defaultBaseDhclientDir
 )
 
 // ipVersion is a wrapper containing the human-readable version string and
@@ -85,7 +90,7 @@ func (n dhclient) Name() string {
 // IsManaging checks if the dhclient CLI is available.
 func (n dhclient) IsManaging(ctx context.Context, iface string) (bool, error) {
 	// Check if the dhclient CLI exists.
-	if _, err := exec.LookPath("dhclient"); err != nil {
+	if _, err := execLookPath("dhclient"); err != nil {
 		if errors.Is(err, exec.ErrNotFound) {
 			return false, nil
 		}
@@ -260,7 +265,9 @@ func dhclientProcessExists(ctx context.Context, iface string, ipVersion ipVersio
 			if ipVersion == ipv6 {
 				return containsProtocolArg, nil
 			}
-			return true, nil
+			if ipVersion == ipv4 && !slices.Contains(commandLine, ipv6.dhclientArg) {
+				return true, nil
+			}
 		}
 	}
 	return false, nil
--- a/google_guest_agent/network/manager/dhclient_linux_test.go
+++ b/google_guest_agent/network/manager/dhclient_linux_test.go
@@ -1,0 +1,511 @@
+//  Copyright 2024 Google LLC
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package manager is responsible for detecting the current network manager service, and
+// writing and rolling back appropriate configurations for each network manager service.
+package manager
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"slices"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/osinfo"
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/ps"
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
+)
+
+// The test DHClient to use in the test.
+var testDHClient = dhclient{}
+
+// The mock Runner client to use for this test.
+type dhclientMockRunner struct {
+	// quietErr indicates whether Quiet() should return an error.
+	quietErr bool
+}
+
+func (d dhclientMockRunner) Quiet(ctx context.Context, name string, args ...string) error {
+	if d.quietErr {
+		// Error every time to see the command being run.
+		var msg = name
+		for _, arg := range args {
+			msg += fmt.Sprintf(" %v", arg)
+		}
+		return fmt.Errorf(msg)
+	}
+	return nil
+}
+
+func (d dhclientMockRunner) WithOutput(ctx context.Context, name string, args ...string) *run.Result {
+	return &run.Result{
+		StdOut: fmt.Sprintf("%v %v", name, args),
+	}
+}
+
+func (d dhclientMockRunner) WithOutputTimeout(ctx context.Context, timeout time.Duration, name string, args ...string) *run.Result {
+	return &run.Result{}
+}
+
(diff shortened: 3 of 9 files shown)


google_guest_agent/network/manager/common.go
```diff
@@ -19,13 +19,20 @@ package manager
...
 
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
 	"github.com/GoogleCloudPlatform/guest-agent/metadata"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
 
 var (
 	badMAC = make(map[string]net.Interface)
+
+	// The following are set for testing purposes.
```
dorileo, line 32: Please add a comment to each of these lines giving enough context to why the exist in the first place.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -58,6 +57,20 @@ var (
 	ipv6 = ipVersion{"ipv6", "-6"}
 )
 
+var (
+	// This next section is for definitions for overriding in tests.
+
+	// baseDhclientDir points to the base directory for DHClient leases and PIDs.
+	baseDhclientDir = defaultBaseDhclientDir
```
dorileo, line 64: I think you are flexing the wrong layer, with these you are bypassing the `runDhclient` and `dhclientProcesExists` and not adding them any coverage.

google_guest_agent/network/manager/dhclient_linux_test.go
```diff
@@ -0,0 +1,318 @@
...
+	testDHClient = dhclient{}
+)
+
+// dhclientTestSetup sets up the test.
+func dhclientTestSetup(t *testing.T, overrideRunDhclient bool) {
+	t.Helper()
+
+	// We have to mock dhclientProcessExists as we cannot mock where the ps
+	// package checks for processes here.
+	dhclientProcessExists = func(ctx context.Context, iface string, ipVersion ipVersion) (bool, error) {
```
dorileo, line 41: Here again we are bypassing a chunk of the code that we want coverage for. You need to "mock" `ps.Find()` instead.

google_guest_agent/network/manager/manager_test.go
```diff
@@ -30,6 +31,9 @@ const (
 	valueFalse = 0
 	valueTrue  = 1
 	valueErr   = 2
+
+	// Version to mock for testing.
```
dorileo, line 35: I can't really understand what this variable is used for by simply reading this comment. Do you mind expanding it a bit?

dorileo (changes requested):

google_guest_agent/network/manager/dhclient_linux.go
drewhli: There is actually coverage for runDhclient. I will add coverage for `dhclientProcessExists`.

The idea behind this is to avoid running those functions and instead run a very dumbed-down version of them for other tests since those tests aren't focused on the functionality of those functions, and we only care about their return values.

google_guest_agent/network/manager/common.go
drewhli: Done.

google_guest_agent/network/manager/dhclient_linux_test.go
drewhli: Added coverage, mocked `ps.Find()`

google_guest_agent/network/manager/common.go
```diff
@@ -19,13 +19,37 @@ package manager
...
+	valueTrue
+	valueErr
+)
+
 var (
 	badMAC = make(map[string]net.Interface)
+
+	// The following are set for testing purposes.
+	// runQuiet points to the function to run commands without output.
+	runQuiet = run.Quiet
```
ChaitanyaKulkarni28, line 42: I would suggest looking at [MDS interface](https://github.com/GoogleCloudPlatform/guest-agent/blob/main/metadata/metadata.go#L53) how its implemented and used for unit testing and see if it makes sense introducing something like that for `ps` or `run`. Problem I see with doing like this is every other test that needs to mock `ps` or `exec` might've to come up with variables like `findProcess` here v/s having consistent interface implementation and just mocking behavior in tests.

drewhli: Created an interface and an overrideable exported variable for overriding the client with a test/mock implementaiton.

google_guest_agent/network/manager/dhclient_linux_test.go
```diff
@@ -0,0 +1,476 @@
...
+				"-r",
+				"iface",
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("test-run-dhclient-%s", test.name), func(t *testing.T) {
+			ctx := context.Background()
+			dhclientTestSetup(t, dhclientTestOpts{})
```
ChaitanyaKulkarni28, line 383: same as above, do we need setup and teardown for each subtest?

google_guest_agent/network/manager/common.go
```diff
@@ -19,13 +19,24 @@ package manager
 import (
 	"fmt"
 	"net"
+	"os/exec"
 
 	"github.com/GoogleCloudPlatform/guest-agent/metadata"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
 
+const (
```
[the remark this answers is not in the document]
drewhli: Yea only used for tests. I've been considering where best to put them, since I'm planning on using them in other tests as well, so I couldn't figure out a good place to put them out of perhaps a new file with purely shared test definitions.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -240,9 +254,9 @@ func partitionInterfaces(ctx context.Context, interfaces, ipv6Interfaces []strin
 	return obtainIpv4Interfaces, obtainIpv6Interfaces, releaseIpv6Interfaces, nil
 }
 
-// dhclientProcessExists checks if a dhclient process for the provided
+// defaultDhclientProcessExists checks if a dhclient process for the provided
 // interface and IP version exists.
-func dhclientProcessExists(ctx context.Context, iface string, ipVersion ipVersion) (bool, error) {
+func defaultDhclientProcessExists(ctx context.Context, iface string, ipVersion ipVersion) (bool, error) {
```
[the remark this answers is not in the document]
drewhli: Done.

google_guest_agent/network/manager/dhclient_linux.go
```diff
@@ -173,9 +187,9 @@ func leaseFilePath(iface string, ipVersion ipVersion) string {
 	return path.Join(baseDhclientDir, leaseFileDir, fmt.Sprintf("dhclient.google-guest-agent.%s.%s.pid", iface, ipVersion.version))
 }
 
-// runDhclient obtains a lease with the provided IP version for the given
+// defaultRunDhclient obtains a lease with the provided IP version for the given
 // network interface. If release is set, this will release leases instead.
-func runDhclient(ctx context.Context, ipVersion ipVersion, nic string, release bool) error {
+func defaultRunDhclient(ctx context.Context, ipVersion ipVersion, nic string, release bool) error {
```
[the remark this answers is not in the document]
drewhli: Using runDhclient() directly now.

google_guest_agent/network/manager/dhclient_linux_test.go
```diff
@@ -0,0 +1,476 @@
...
+
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("test-partition-interfaces-%s", test.name), func(t *testing.T) {
+			ctx := context.Background()
+			opts := dhclientTestOpts{
+				processOpts: dhclientProcessOpts{
+					override: true,
+				},
+			}
+			dhclientTestSetup(t, opts)
```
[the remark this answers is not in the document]
drewhli: Moved out for tests that don't require them. There was some refactoring to make using the original functions work.

google_guest_agent/network/manager/dhclient_linux_test.go
```diff
@@ -0,0 +1,476 @@
...
+
+			obtainIpv4, obtainIpv6, releaseIpv6, err := partitionInterfaces(ctx, test.testInterfaces, test.testIpv6Interfaces)
+			if err != nil {
+				t.Fatalf("partitionInterfaces return error when none expected: %v", err)
+			}
+			if !slices.Equal(obtainIpv4, test.expectedObtainIpv4) {
+				t.Errorf("partitionInterfaces did not return correct obtainIpv4. Expected: %v, Actual: %v", test.expectedObtainIpv4, obtainIpv4)
```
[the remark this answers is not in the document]
drewhli: Done.

google_guest_agent/network/manager/dhclient_linux_test.go
drewhli: Only for some, not all, post-refactor.

google_guest_agent/network/manager/common.go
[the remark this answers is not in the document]
dorileo: Do we need it in the first place? Can we consider using booleans for that purpose? My feeling is that if we need to define an enum to check a test result we are probably doing it wrong or over complicating the test implementation. Would breaking the test into two or three different test functions resolve or make it simpler?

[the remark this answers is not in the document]
drewhli: The enum was mainly to avoid having to declare 2 sets of variables -- one to tell the mock functions to return an error, and one to tell the mock function to return true/false. Breaking the test down would be a good idea, though. I've opted for an extra bool value instead.

google_guest_agent/network/manager/dhclient_linux_test.go
```diff
@@ -0,0 +1,511 @@
...
+			findPath: func(path string) (string, error) {
+				return "", fmt.Errorf("mock error")
+			},
+			expectedBool: false,
+			expectErr:    true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(fmt.Sprintf("test-is-managing-%s", test.name), func(t *testing.T) {
```
ChaitanyaKulkarni28, line 226: nit: `test-is-managing` is redundant here (and all other usages) as all subtests would be named `<function-name>/<sub-test-name>`

PR #355, opened 2024-02-12
Refactor guest-agent to use common retry util

ChaitanyaKulkarni28:


google_guest_agent/instance_setup.go
```diff
@@ -95,12 +97,10 @@ func agentInit(ctx context.Context) {
 
 	if runtime.GOOS == "windows" {
 		// Indefinitely retry to set up required MDS route.
```
dorileo, line 99: We, in this case we not Indefinitely retrying, so this (although a huge limit). I wonder if we even want to keep trying that long, wouldn't we be hiding actual issues by trying "forever"?

ChaitanyaKulkarni28: not sure what limit should be, I kept it as is assuming service startup limit would kick in and fail anyways if this limit is too high. Any recommendation on limit if we want to lower it?

ChaitanyaKulkarni28: Changed it to limit retries for max 1 min.

PR #356, opened 2024-02-13
Create wicked network service implementation and its respective unit tests

drewhli:
This also refactors the `options` structs in the `systemd-networkd` tests to avoid type overlap.

/hold

### DIFF
--- a/google_guest_agent/network/manager/wicked_linux.go
+++ b/google_guest_agent/network/manager/wicked_linux.go
@@ -1,0 +1,181 @@
+//  Copyright 2024 Google LLC
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package manager is responsible for detecting the current network manager service, and
+// writing and rolling back appropriate configurations for each network manager service.
+package manager
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/cfg"
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+)
+
+type wicked struct {
+	// configDir is the directory to which to write configuration files.
+	configDir string
+}
+
+const (
+	// defaultWickedConfigDir is the default location for wicked configuration files.
+	defaultWickedConfigDir = "/etc/sysconfig/network"
+
+	// wickedCommand is the expected path to the wicked CLI.
+	wickedCommand = "/usr/sbin/wicked"
+)
+
+// init adds this network manager service to the list of known managers.
+func init() {
+	registerManager(&wicked{
+		configDir: defaultWickedConfigDir,
+	}, false)
+}
+
+// Name returns the name of this network manager service.
+func (n wicked) Name() string {
+	return "wicked"
+}
+
+// IsManaging checks whether wicked is managing the provided interface.
+func (n wicked) IsManaging(ctx context.Context, iface string) (bool, error) {
+	// Check the current main network service. Primarily applicable to SUSE images.
+	res := run.WithOutput(ctx, "systemctl", "status", "network.service")
+	if strings.Contains(res.StdOut, "wicked.service") {
+		return true, nil
+	}
+
+	// Check if the wicked service is running.
+	res = run.WithOutput(ctx, "systemctl", "is-active", "wicked.service")
+	if res.ExitCode != 0 {
+		return false, nil
+	}
+
+	// Check the status of configured interfaces.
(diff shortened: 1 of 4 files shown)


google_guest_agent/network/manager/wicked_linux.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+type wicked struct {
+	// configDir is the directory to which to write configuration files.
+	configDir string
+}
+
+const (
+	// defaultWickedConfigDir is the default location for wicked configuration files.
+	defaultWickedConfigDir = "/etc/sysconfig/network"
```
dorileo, line 39: Does wicd also look into `/usr/lib/`? I'd favor owning files there and leaving `/etc/sysconfig/` to user defined configs .

google_guest_agent/network/manager/wicked_linux.go
```diff
@@ -0,0 +1,178 @@
...
+		ifcfg, err := os.Create(ifcfgFilePath(configDir, iface))
+		if err != nil {
+			return err
+		}
+		defer ifcfg.Close()
+		contents := []string{
+			googleComment,
+			"STARTMODE=hotplug",
+			// NOTE: 'dhcp' is the dhcp4+dhcp6 option.
+			"BOOTPROTO=dhcp",
```
dorileo, line 121: do we want to make it compatible/consistent with `systemd-networkd` and set dhcp4 or dhcp6 or dhcp depending on the network stack configuration provided by mds?

google_guest_agent/network/manager/wicked_linux.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	args := append([]string{"ifup", "--timeout", "1"}, ifaces[1:]...)
+	return run.Quiet(ctx, wickedCommand, args...)
+}
+
+// writeWickedConfigs writes config files for the given ifaces in the given configuration
+// directory.
+func writeWickedConfigs(configDir string, ifaces []string) error {
+	var priority = 10100
```
dorileo, line 104: do we know why we are setting priority to 10100?

google_guest_agent/network/manager/wicked_linux.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// Write the config for all the non-primary network interfaces.
+	for _, iface := range ifaces {
+		logger.Debugf("write enabling ifcfg-%s config", iface)
```
dorileo, line 108: "writing wicked's if-cfg-%s config"

google_guest_agent/network/manager/wicked_linux.go
```diff
@@ -0,0 +1,178 @@
...
+		contents := []string{
+			googleComment,
+			"STARTMODE=hotplug",
+			// NOTE: 'dhcp' is the dhcp4+dhcp6 option.
+			"BOOTPROTO=dhcp",
+			fmt.Sprintf("DHCLIENT_ROUTE_PRIORITY=%d", priority),
+		}
+		_, err = ifcfg.WriteString(strings.Join(contents, "\n"))
+		if err != nil {
+			return err
```
dorileo, line 126: please wrap this WriteString error.

google_guest_agent/network/manager/wicked_linux.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		priority += 100
```
dorileo, line 128: why do we need incremented priority per interface again?

google_guest_agent/network/manager/wicked_linux.go
```diff
@@ -0,0 +1,178 @@
...
+// Rollback deletes all the ifcfg files written by Setup, then reloads wicked.service.
+func (n wicked) Rollback(ctx context.Context, payload []metadata.NetworkInterfaces) error {
+	ifaces, err := interfaceNames(payload)
+	if err != nil {
+		return fmt.Errorf("failed to get network interfaces: %v", err)
+	}
+
+	// Since configuration files are only written for non-primary, only check
+	// for non-primary configuration files.
+	for _, iface := range ifaces[1:] {
```
dorileo, line 147: Do we want to rollback only secondary nics? It seems that we are writing configs for primary as well.

google_guest_agent/network/manager/wicked_linux.go
drewhli: No, it only seems to write to `/etc/sysconfig/`

google_guest_agent/network/manager/wicked_linux.go
drewhli: I've been snooping around some documentation, but I haven't found anything concrete. But in [this example](https://www.suse.com/support/kb/doc/?id=000021237), they increment the priority by 100 past the first interface. This is mainly to maintain parity with the old implementation, but I agree that it seems pretty arbitrary at the moment.

google_guest_agent/network/manager/wicked_linux.go
drewhli: I honestly do not really know. It seems that the field is used for DHCPv4 route priorities, though it's unclear whether higher priority means higher in order or lower in order.

google_guest_agent/network/manager/wicked_linux.go
drewhli: Edit: the BOOTPROTO field only supports `none`, `bootp`, and `dhcp`, so keeping it the same.

google_guest_agent/network/manager/wicked_linux.go
dorileo: Indeed.

google_guest_agent/network/manager/wicked_linux.go
dorileo: Based on my understanding we want only to have eth0 < eth1 < ethN. It seems to be common practice to increase by 100 but to me is looks more like a range reservation rather than anything else. It also connects with the initial value of 10100, it could be 100 or 1.

dorileo (approved):

PR #357, opened 2024-02-14
Update metadata script runner to use retry util, add unit tests

ChaitanyaKulkarni28:


google_metadata_script_runner/main.go
```diff
@@ -103,36 +106,35 @@ func downloadGSURL(ctx context.Context, bucket, object string, file *os.File) er
 	}
 	defer client.Close()
 
-	r, err := client.Bucket(bucket).Object(object).NewReader(ctx)
+	r, err := retry.RunWithResponse(ctx, defaultRetryPolicy, func() (*storage.Reader, error) {
+		r, err := client.Bucket(bucket).Object(object).NewReader(ctx)
+		return r, err
```
dorileo, line 111: don't we want to wrap the error in here?

dorileo (changes requested):
Just a minor comment/request.

google_metadata_script_runner/main.go
ChaitanyaKulkarni28: not required, as caller already knows what bucket and object its downloading at its logged on line 160, so to avoid duplication I sent it as is.

PR #362, opened 2024-02-20
Create NetworkManager implementation

drewhli:
- Deleted the `nativeOSConfig` field in the `osRules` as that will no longer be needed.
- Deleted the `broadVersion` and related variables, as they will no longer be used after this and the Wicked implementation.
- Created NetworkManager tests.

The current behavior is as follows:
1. Guest Agent writes its own `.nmconnections` file for each interface under `/etc/NetworkManager/system-connections`.
2. We then load those connection files into `nmcli`, then enable those new connections
  i. This overrides the default system connections, which are called `Wired Connection 1`, `Wired Connection 2`, etc...
3. Upon rebooting the system, we lose the default system connections. This is more or less intended because those connections reside in the `/run` directory, which gets cleared upon reboot.

### DIFF
--- a/google_guest_agent/network/manager/network_manager_linux.go
+++ b/google_guest_agent/network/manager/network_manager_linux.go
@@ -1,0 +1,247 @@
+//  Copyright 2024 Google LLC
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package manager is responsible for detecting the current network manager service, and
+// writing and rolling back appropriate configurations for each network manager service.
+package manager
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/cfg"
+	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
+	"github.com/GoogleCloudPlatform/guest-agent/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	"github.com/go-ini/ini"
+)
+
+const (
+	defaultNetworkManagerConfigDir = "/etc/NetworkManager/system-connections"
+)
+
+// nmConnectionSection is the connection section of NetworkManager's keyfile.
+type nmConnectionSection struct {
+	// InterfaceName is the name of the interface to configure.
+	InterfaceName string `ini:"interface-name"`
+
+	// ID is the unique ID for this connection.
+	ID string `ini:"id"`
+
+	// ConnType is the type of connection (i.e. ethernet).
+	ConnType string `ini:"type"`
+}
+
+// nmIpv4Section is the ipv4 section of NetworkManager's keyfile.
+type nmIpv4Section struct {
+	// Method is the IP configuration method. Supports "auto", "manual", and "link-local".
+	Method string
+}
+
+// nmIpv6Section is the ipv6 section of NetworkManager's keyfile.
+type nmIpv6Section struct {
+	// Method is the IP configuration method. Supports "auto", "manual", and "link-local".
+	Method string
+}
+
+// nmConfig is a wrapper containing all the sections for the NetworkManager keyfile.
+type nmConfig struct {
+	// GuestAgent is the 'guest-agent' section.
+	GuestAgent guestAgentSection `ini:"guest-agent"`
+
+	// Connection is the connection section.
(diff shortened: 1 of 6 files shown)


drewhli:
/hold

Just realized I forgot to implement one of the tests.

google_guest_agent/network/manager/network_manager_linux.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/go-ini/ini"
+)
+
+const (
+	defaultNetworkManagerConfigDir = "/etc/NetworkManager/system-connections"
+)
+
+// nmGuestAgentSection is a custom section to indicate that this connection is
+// managed by the guest agent.
+type nmGuestAgentSection struct {
```
dorileo, line 41: you can move systemd-networkd's `guestAgent` struct and reuse it here, move it up to `manager.go`, they look the same.

google_guest_agent/network/manager/network_manager_linux.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// Name is the name of this network manager service.
+func (n networkManager) Name() string {
+	return "NetworkManager"
+}
+
+// IsManaging checks if NetworkManager is managing the provided interface.
+func (n networkManager) IsManaging(ctx context.Context, iface string) (bool, error) {
+	// Check the status of network.service. This mainly works on SUSE distributions.
```
dorileo, line 105: I don't get the `This mainly works on SUSE distributions` comment here.

google_guest_agent/network/manager/network_manager_linux.go
```diff
@@ -0,0 +1,268 @@
...
+	res := run.WithOutput(ctx, "systemctl", "status", "network.service")
+	if strings.Contains(res.StdOut, "NetworkManager.service") {
```
dorileo, line 107: this will only be true if the user is running `NetworkManager` on suse, is that the expected?

google_guest_agent/network/manager/network_manager_linux.go
```diff
@@ -0,0 +1,268 @@
...
+		return false, fmt.Errorf("error checking for nmcli: %v", err)
+	}
+
+	// Check whether NetworkManager.service is active.
+	if err := run.Quiet(ctx, "systemctl", "is-active", "NetworkManager.service"); err != nil {
+		return false, nil
+	}
+
+	// Use nmcli to check status of provided  interface.
+	res = run.WithOutput(ctx, "nmcli", "-f", "DEVICE,STATE", "dev", "status")
```
dorileo, line 126: if you use `-t` you'll get a result CSV like output, a more suitable for computers reading format (ps: it can be combined `-f DEVICE,STATE`).

google_guest_agent/network/manager/network_manager_linux.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		return fmt.Errorf("error writing NetworkManager connection configs: %v", err)
+	}
+
+	// Load all the connections.
+	for _, iface := range ifaces {
+		configFile := n.networkManagerConfigFilePath(iface)
+
+		// NetworkManager can only load and use config files with 600 permission.
+		if err := os.Chmod(configFile, 0600); err != nil {
```
dorileo, line 159: I'd rather have this `Chmod()` call closer to where the file is created i.e. `writeNetworkManagerConfigs()`.

google_guest_agent/network/manager/network_manager_linux.go
```diff
@@ -0,0 +1,268 @@
...
+			return fmt.Errorf("error updating permissions for %s: %v", configFile, err)
+		}
+		if err = run.Quiet(ctx, "nmcli", "conn", "load", configFile); err != nil {
```
dorileo, line 162: Are you doing this `nmcli conn load` in this level rather than in the `writeNetworkManagerConfigs()` to process all of them after you've written them all? considering that we area already doing a `nmcli conn reload` after this block I'd assume that it's basically doing the same thing again, is it not the case?

dorileo (changes requested):

google_guest_agent/network/manager/network_manager_linux.go
drewhli: On SUSE distributions, you can check the status of `network.service` to see which network manager service is managing the networking on the VM. This is basically a shortcut on SUSE distros for checking the primary network manager.

google_guest_agent/network/manager/network_manager_linux.go
drewhli: No, typically SUSE distributions *should* be using `wicked` as far as I know, but this would account for the "in-case" factor.

google_guest_agent/network/manager/network_manager_linux.go
drewhli: Noted. Will change.

google_guest_agent/network/manager/network_manager_linux.go
drewhli: I was testing this without doing the `chmod 600` thing before, which caused `nmcli conn reload` to not load the new guest-agent-created files. Tested just now, and it seems like we don't need the `load` command anymore. Will remove, thanks.

google_guest_agent/network/manager/network_manager_linux.go
drewhli: Done.

dorileo (approved):

PR #365, opened 2024-02-26
Add COS homedir-gid patch to upstream.

mike-kochera:
This patch was made to fix a problem where you can become unable to create a user due to a UID/GID conflict. We've maintained the patch but it'd be better to upstream it so we don't have to worry about a regression while upgrading the guest agent in COS.

### DIFF
--- a/google_guest_agent/accounts_unix.go
+++ b/google_guest_agent/accounts_unix.go
@@ -27,20 +27,29 @@ import (
 	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/run"
 )
 
-func getUID(path string) string {
+func getUIDAndGID(path string) (string, string) {
 	if dir, err := os.Stat(path); err == nil {
 		if stat, ok := dir.Sys().(*syscall.Stat_t); ok {
-			return fmt.Sprintf("%d", stat.Uid)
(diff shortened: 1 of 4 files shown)


google_guest_agent/accounts_unix.go
```diff
@@ -27,21 +27,30 @@ import (
...
 	config := cfg.Get()
 	useradd := config.Accounts.UserAddCmd
 	if uid != "" {
 		useradd = fmt.Sprintf("%s -u %s", useradd, uid)
 	}
+	if gid != "" {
+		groupadd := config.Accounts.GroupAddCmd
+		groupadd = fmt.Sprintf("%s -g %s", groupadd, gid)
+		cmd, args := createUserGroupCmd(groupadd, "", username)
+		if err := run.Quiet(ctx, cmd, args...); err != nil {
```
ChaitanyaKulkarni28, line 49: will there be a case where group already exists? as probably this will fail if group already exists? I believe we want to continue in that case and not return error.

dexter-1: I don't think there would be a case where the group with name `username` already exists because this block of code will only run if the user does not exist on the system. 

However, the `groupadd` command could return an error if the GID already exists. In that case, we would want to return the error since this indicates that some other group already claimed the GID so we can't create the user with that GID.

ChaitanyaKulkarni28 (approved):

PR #368, opened 2024-03-07
Add require-oslogin-certificates logic to disable keys

bpl4vv:
Add logic to read `require-oslogin-certificates` from metadata and disable keys if true

### DIFF
--- a/google_guest_agent/oslogin.go
+++ b/google_guest_agent/oslogin.go
@@ -75,7 +75,7 @@ func getOSLoginEnabled(md *metadata.Descriptor) (bool, bool, bool, bool) {
 	}
 	var reqCerts bool
 	if md.Project.Attributes.RequireCerts != nil {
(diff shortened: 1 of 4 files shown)


google_guest_agent/oslogin.go
```diff
@@ -312,12 +320,15 @@ func updateSSHConfig(sshConfig string, enable, twofactor, skey bool) string {
 	if enable {
 		osLoginBlock := []string{googleBlockStart}
 
-		if cfg.Get().OSLogin.CertAuthentication {
+		// Metadata overrides the config file.
```
ChaitanyaKulkarni28, line 323: I guess this can be simplified by - 

```
if  reqCerts ||  cfg.Get().OSLogin.CertAuthentication {
 ---
}

dorileo (approved):

PR #395, opened 2024-05-23
netplan: add support for default MTU value from mds

dorileo:
MDS will provide us with the user's defined MTU make sure we honor it on the netplan dropin.


dorileo:
It actually doesn't fix the issue but I managed to reproduce the problem and a solution (for the Martian packets issue at least).

PR #396, opened 2024-06-08
Network subsystem remove os rules

dorileo:


google_guest_agent/network/manager/manager_linux.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func init() {
+	wickedCommand, err := exec.LookPath("wicked")
+	if err != nil {
+		logger.Infof("failed to find wicked path, falling back to default: %+v", err)
+		wickedCommand = defaultWickedCommand
+	}
+
+	knownNetworkManagers = []Service{
```
a-crate, line 30: I think setting an already declared var in `init` to a value that's more or less static is needlessly confusing. IMO this slice should be set as part of the declaration and `init()` in `wicked_linux.go` should edit wickedCommand if exec.LookPath is successful.

google_guest_agent/network/manager/netplan_linux.go
```diff
@@ -244,7 +268,7 @@ func (nd networkdNetplanDropin) write(n netplan, iface string) error {
 	logger.Infof("writing systemd drop in to: %s", dropinFile)
 
 	dropinDir := filepath.Dir(dropinFile)
-	err := os.MkdirAll(dropinDir, 0750)
+	err := os.MkdirAll(dropinDir, 0755)
```
a-crate, line 271: Why did this change?

google_guest_agent/network/manager/netplan_linux.go
```diff
@@ -88,11 +79,22 @@ type netplanEthernet struct {
...
+	DHCP4Overrides *netplanDHCPOverrides `yaml:"dhcp4-overrides,omitempty"`
+
 	// DHCPv6 determines if DHCPv6 support must be enabled to such an interface.
 	DHCPv6 *bool `yaml:"dhcp6,omitempty"`
+
+	DHCP6Overrides *netplanDHCPOverrides `yaml:"dhcp6-overrides,omitempty"`
+}
+
+type netplanDHCPOverrides struct {
+	UseDomains bool `yaml:"use-domains,omitempty"`
```
a-crate, line 97: Can you add a comment about using this to add a search domain for whatever DNS domain the instance is in

a-crate (commented):
lgtm

google_guest_agent/network/manager/netplan_linux.go
dorileo: it was wrong in the first place.

google_guest_agent/network/manager/netplan_linux.go
dorileo: done. thanks.

a-crate (approved):

PR #400, opened 2024-07-04
Disable OS Login without pruning off any extra suffix.

ericdand:
This small change fixes an issue where, if a user had added more NSS plugins to their `/etc/nsswitch.conf` file, disabling OS Login would cause those extra plugins to be silently dropped from the NSS config. 

Although we don't strictly support using OS Login with other NSS plugins, we should make sure that our installation script isn't so brittle that it quietly breaks other NSS plugins.


a-crate:
Can you add a unit test case for this in `TestUpdateNSSSwitchConfig()`

ericdand:
Done! :) PTAL.

a-crate (approved):

PR #402, opened 2024-07-17
Update NIC management strategy

a-crate:
See commit messages for details. Short version: bring back OS rules with a configuration option to ignore them. 
Stop removing the default configuration on Debian 12, write a config with a higher priority.
Add a network recovery strategy, executed when then MDS is inaccessible on startup. Remove all guest agent configuration and go back to the OS default, exit with a failure if the MDS is still inaccessible with the default configuration.

/hold
/cc @ChaitanyaKulkarni28 @drewhli

PR #408, opened 2024-07-31
Configure primary nic if only set in cfg file

ChaitanyaKulkarni28:
This is follow up PR for #402 to make sure use-domains is set as expected and primary nic configuration is dependent on config value set in the file

/cc @a-crate @drewhli

### DIFF
--- a/google_guest_agent/network/manager/netplan_linux.go
+++ b/google_guest_agent/network/manager/netplan_linux.go
@@ -97,7 +97,7 @@ type netplanEthernet struct {
 type netplanDHCPOverrides struct {
 	// When true, the domain name received from the DHCP server will be used as DNS
 	// search domain over this link.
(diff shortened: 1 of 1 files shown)


google_guest_agent/network/manager/netplan_linux.go
```diff
@@ -274,6 +277,16 @@ func (nd networkdNetplanDropin) write(n netplan, iface string) error {
 	return nil
 }
 
+// shouldUseDomains returns true if interface index is 0.
+func shouldUseDomains(idx int) *bool {
+	trueVal := true
```
a-crate, line 282: This can be simpler, can we just do something like `val := idx ==0; return &val`?

ChaitanyaKulkarni28: Done

drewhli (approved):

PR #410, opened 2024-08-02
Avoid writing configuration files when they already exist on wicked and NetworkManager

drewhli:
The `wicked` change is an upstream from an [existing patch](https://build.opensuse.org/projects/openSUSE:Factory/packages/google-guest-agent/files/dont_overwrite_ifcfg.patch?expand=1 ). We also do the same for `NetworkManager` since they use the same `ifcfg` files, and no priority system exists for those. In this case, we only delete the old configuration files if they are managed by the guest agent. These are files written by the previous implementation of the network stack that simply disabled `NetworkManager`.

Included in this also are some small comment nits for `netplan`.

/cc @dorileo @a-crate @ChaitanyaKulkarni28

### DIFF
--- a/google_guest_agent/network/manager/network_manager_linux.go
+++ b/google_guest_agent/network/manager/network_manager_linux.go
@@ -204,14 +204,17 @@ func (n networkManager) writeNetworkManagerConfigs(ifaces []string) ([]string, e
 			return []string{}, fmt.Errorf("error updating permissions for %s connection config: %v", iface, err)
 		}
 
+		// Clean up the files written by the old agent. Make sure they're managed
+		// by the agent before deleting them.
 		ifcfgFilePath := n.ifcfgFilePath(iface)
-		_, err := os.Stat(ifcfgFilePath)
-		if err != nil {
-			if !os.IsNotExist(err) {
-				return nil, fmt.Errorf("failed to stat ifcfg file(%s): %v", ifcfgFilePath, err)
-			}
-		} else {
-			if err := os.Remove(ifcfgFilePath); err != nil {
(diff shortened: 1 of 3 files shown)


google_guest_agent/network/manager/network_manager_linux.go
```diff
@@ -204,14 +205,17 @@ func (n networkManager) writeNetworkManagerConfigs(ifaces []string) ([]string, e
...
-			}
-		} else {
-			if err := os.Remove(ifcfgFilePath); err != nil {
+		contents, err := os.ReadFile(ifcfgFilePath)
+		if err != nil && !os.IsNotExist(err) {
+			return nil, fmt.Errorf("failed to read ifcfg file(%s): %v", ifcfgFilePath, err)
+		}
+
+		// Check for the google comment.
+		if slices.Contains(strings.Split(string(contents), "\n"), "# Added by Google Compute Engine OS Login.") {
```
ChaitanyaKulkarni28, line 217: do we care about comment being a separate line and all or can we do `strings.Contains(string(contents), "....")` directly?

dorileo:
do we need to have it ported to stable branch?

google_guest_agent/network/manager/network_manager_linux.go
drewhli: We can check directly. Done.

ChaitanyaKulkarni28:
> do we need to have it ported to stable branch?

I don't think any of network manager changes are in stable

dorileo:
> > do we need to have it ported to stable branch?
> 
> I don't think any of network manager changes are in stable

indeed. Thanks for checking.

PR #411, opened 2024-08-05
Get rid of deprecated dependencies in snapshot service 

ChaitanyaKulkarni28:
Snapshot service is using deprecated `github.com/golang/protobuf`, update it to use latest `google.golang.org/protobuf`


dorileo:
lgtm

PR #413, opened 2024-08-05
Introduce a configuration toggle for enabling/disabling cloud logging

ChaitanyaKulkarni28:
Guest Agent had by default cloud logging enabled with no way to disable it. As long as permissions are set agent would log into cloud logging. 

Users might want to disable any external access including google apis, allow disabling it without having to play with permissions, disabling the logging api or introduce firewall rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ Accounts          | usermod\_cmd           | Command string to modify a user's g
 Accounts          | gpasswd\_add\_cmd      | Command string to add a user to a group.
 Accounts          | gpasswd\_remove\_cmd   | Command string to remove a user from a group.
 Accounts          | groupadd\_cmd          | Command string to create a new group.
+Core              | cloud\_logging\_enabled| `false` disable cloud logging.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -196,6 +196,7 @@ Accounts          | usermod\_cmd           | Command string to modify a user's g
 Accounts          | gpasswd\_add\_cmd      | Command string to add a user to a group.
 Accounts          | gpasswd\_remove\_cmd   | Command string to remove a user from a group.
 Accounts          | groupadd\_cmd          | Command string to create a new group.
+Core              | cloud\_logging\_enabled| `false` disable cloud logging.
```
dorileo, line 199: it seems the default config is setting it to true, isn't it?

ChaitanyaKulkarni28: yes, but entries here do not define default values. Like `accounts_daemon` is [true](https://github.com/GoogleCloudPlatform/guest-agent/blob/main/google_guest_agent/cfg/cfg.go#L55) by default

dorileo: indeed.

PR #414, opened 2024-08-06
Add a configuration option to disable CA certificate updates

ChaitanyaKulkarni28:
Agent is writing certs on disk along with updating the root trust store, add a configuration toggle to disable system cacerts update instead of having to disable whole feature. With this users can still use the feature by using certs from `/run/google-mds-mtls`


ChaitanyaKulkarni28:
Fixed with https://github.com/GoogleCloudPlatform/guest-agent/pull/423

PR #426, opened 2024-09-03
Adapt and update mtls mds config key to be consistent with MDS

ChaitanyaKulkarni28:
Replaces key name from `https-mds-skip-native-cert-store` to `enable-https-mds-native-cert-store` to be consistent with MDS.

/cc @dorileo @drewhli @a-crate


google_guest_agent/agentcrypto/mtls_mds.go
```diff
@@ -267,17 +267,17 @@ func (j *CredsJob) shouldUseNativeStore(mds *metadata.Descriptor) bool {
 	var useNative bool
 
 	if cfg.Get().MDS != nil {
-		useNative = !cfg.Get().MDS.HTTPSMDSSkipNativeStore
+		useNative = cfg.Get().MDS.HTTPSMDSEnableNativeStore
```
a-crate, line 270: Is the VM's config file not allowed to enable the native cert store if it's disabled in metadata somewhere?

ChaitanyaKulkarni28: yes, metadata takes higher priority and allows users to configure without being on VM or restarting agent and at scale (can simply set it at project level).

PR #427, opened 2024-09-09
Update README to reflect new network manager changes

drewhli:

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,15 +108,68 @@ skew may result in `system time has changed` messages in VM logs.
 #### Network
 
 The guest agent uses network interface metadata to manage the network
-interfaces in the guest by performing the following tasks:
+interfaces by performing the following tasks:
 
-*   Enabled all associated network interfaces on boot.
+*   Enable all associated network interfaces on boot.
+    *   Detect the current active network manager service that is managing the
+        primary NIC.
+    *   Rollback and delete any guest agent-managed files/configurations left
+        behind by all other supported network manager services.
+    *   Write and apply new configurations for the secondary NICs using the
+        network manager service detected in the first step.
+    *   Create a route to the metadata server for the primary NIC only.
 *   Setup or remove IP routes in the guest for IP forwarding and IP aliases
     *   Only IPv4 IP addresses are currently supported.
     *   Routes are set on the primary ethernet interface.
     *   Google routes are configured, by default, with the routing protocol ID
         `66`. This ID is a namespace for daemon configured IP addresses. It can
         be changed with the config file, see below.
+
+On Linux, supported network managers are as follows. These are listed by
+descending priority and include the location at which the configuration files
+are written.
+
+*   `netplan`
+    *   Config location: `/run/netplan/`
+        *   ex: `/run/netplan/20-google-guest-agent-eth0.yaml`
+    *   Dropin location: `/etc/systemd/network/`
+        *   ex: `/etc/systemd/network/10-netplan-eth0.network.d/`
+*   `wicked`
+    *   Config location: `/etc/sysconfig/network/`
+        *   ex: `/etc/sysconfig/network/ifcfg-eth0`
+    *   Notes:
+        *   Existing `ifcfg` files are not overwritten and are skipped instead.
+*   `NetworkManager`
+    *   Config location: `/etc/NetworkManager/system-connections/`
+        *   ex:
+            `/etc/NetworkManager/system-connections/google-guest-agent-eth0.nmconnection`
+*   `systemd-networkd`
+    *   Config location: `/usr/lib/systemd/network/`
+        *   ex: `/usr/lib/systemd/network/20-eth0-google-guest-agent.network`
+*   `dhclient`
+    *   Config location: `/run/`
+        *   ex (pid):   `/run/dhclient.google-guest-agent.eth0.ipv4.pid`
+        *   ex (lease): `/run/dhclient.google-guest-agent.eth0.ipv4.lease`
+    *   Notes:
+        *   The primary NIC setup, if enabled, is skipped if a dhclient process
+            for the primary NIC is already running.
+
+If none of the first 4 network manager services are detected on the system, then
+the agent will default to using `dhclient` for managing network interfaces.
+
+The following configuration flags can control the behavior:
+
+*   `manage_primary_nic`: When enabled, the agent will start managing the
+    primary NIC in addition to the secondary NICs.
+
+For more information about the instance configuration, see the Configuration
+section.
+
+The guest agent will also setup VLANs if VLAN is enabled. The setup and
+configuration for this work similarly to the normal NIC configuration.
+
+If the VLANs' parent interface is the primary NIC, it will apply the VLAN
+configurations regardless of whether `manage_primary_nic` is set.
 
 #### Windows Failover Cluster Support
 


README.md
```diff
@@ -108,16 +108,37 @@ skew may result in `system time has changed` messages in VM logs.
...
 *   Setup or remove IP routes in the guest for IP forwarding and IP aliases
     *   Only IPv4 IP addresses are currently supported.
     *   Routes are set on the primary ethernet interface.
     *   Google routes are configured, by default, with the routing protocol ID
         `66`. This ID is a namespace for daemon configured IP addresses. It can
         be changed with the config file, see below.
 
+On Linux, supported network managers are:
+
+*   `systemd-networkd`
```
bkatyl, line 129: Shouldn't dhclient be in this list since the agent will default to using it if no other is present?

Does this list need to have a priority over as the way dhclient is stated it sounds like it's the fallback.

README.md
```diff
@@ -108,16 +108,37 @@ skew may result in `system time has changed` messages in VM logs.
...
+*   `wicked`
+*   `NetworkManager`
+*   `netplan`
+
+If none of these network interfaces are detected on the system, then the agent
+will default to using `dhclient` for managing network interfaces. The agent will
+not manage the primary NIC unless `manage_primary_nic` is set to `true` in the
```
bkatyl, line 136:
```
The agent will
not manage the primary NIC unless `manage_primary_nic` is set to `true` in the
instance configuration.
```

Is this statement in the context of dhclient or is it a univeral statement?

If it's a universal statement this I'd suggest using a non-paragraph approach to each that each statement can be read as standing alone and of equal merit.

I'd also suggest that "the following features depend on `manage_primary_nic` being enabled. 
* feature abc: This enables foo, etc.

README.md
drewhli: I've updated it to reflect the priority.

README.md
drewhli: Non-paragraph-ized this section and made the blurb about the configuration flag more clear.

README.md
```diff
@@ -108,16 +108,44 @@ skew may result in `system time has changed` messages in VM logs.
...
+    *   Write and apply new configurations for the secondary NICs using the
+        network manager service detected in the first step.
 *   Setup or remove IP routes in the guest for IP forwarding and IP aliases
     *   Only IPv4 IP addresses are currently supported.
     *   Routes are set on the primary ethernet interface.
     *   Google routes are configured, by default, with the routing protocol ID
         `66`. This ID is a namespace for daemon configured IP addresses. It can
         be changed with the config file, see below.
 
+On Linux, supported network managers are (listed by priority):
```
ChaitanyaKulkarni28, line 127: should we also document the paths and priority it uses to make it easier to read. 

I am thinking of taking an example of `eth0` as interface and writing down where to expect config files in each case. 

For ex. in case of netplan we write `/run/netplan/20-google-guest-agent-ethernet.yaml` it will be good to have explicit paths here rather than having to go through code to figure out

README.md
```diff
@@ -108,16 +108,44 @@ skew may result in `system time has changed` messages in VM logs.
...
+
+The following configuration flags can control the behavior:
+
+*   `manage_primary_nic`: When enabled, the agent will start managing the
+    primary NIC in addition to the secondary NICs.
+
+For more information about the instance configuration, see the Configuration
+section.
+
+The guest agent will also setup VLANs if VLAN is enabled. The setup and
```
ChaitanyaKulkarni28, line 146: can you also add a note here about agent applying config for vlan nic whose parent is primary nic regardless of `manage_primary_nic` entry

README.md
```diff
@@ -108,16 +108,63 @@ skew may result in `system time has changed` messages in VM logs.
...
+*   `wicked`
+    *   Config location: `/etc/sysconfig/wicked`
+        *   ex: `/etc/sysconfig/ifcfg-eth0`
+*   `NetworkManager`
+    *   Config location: `/etc/NetworkManager/system-connections/`
+        *   ex:
+            `/etc/NetworkManager/system-connections/google-guest-agent-eth0.nmconnection`
+*   `systemd-networkd`
+    *   Config location: `/usr/lib/systemd/network/`
+        *   ex: /usr/lib/systemd/network/20-eth0-google-guest-agent.network`
```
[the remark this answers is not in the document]
drewhli: Done.

README.md
drewhli: Done. Added the paths and priorities.

README.md
drewhli: Done.

README.md
```diff
@@ -108,16 +108,63 @@ skew may result in `system time has changed` messages in VM logs.
...
+    *   Write and apply new configurations for the secondary NICs using the
+        network manager service detected in the first step.
 *   Setup or remove IP routes in the guest for IP forwarding and IP aliases
     *   Only IPv4 IP addresses are currently supported.
     *   Routes are set on the primary ethernet interface.
     *   Google routes are configured, by default, with the routing protocol ID
         `66`. This ID is a namespace for daemon configured IP addresses. It can
         be changed with the config file, see below.
 
+On Linux, supported network managers are as follows. These are listed by
```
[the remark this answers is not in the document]
drewhli: Added some notes about the caveats.

README.md
```diff
@@ -108,16 +108,68 @@ skew may result in `system time has changed` messages in VM logs.
...
+descending priority and include the location at which the configuration files
+are written.
+
+*   `netplan`
+    *   Config location: `/run/netplan/`
+        *   ex: `/run/netplan/20-google-guest-agent-eth0.yaml`
+    *   Dropin location: `/etc/systemd/network/`
+        *   ex: `/etc/systemd/network/10-netplan-eth0.network.d/`
+*   `wicked`
+    *   Config location: `/etc/sysconfig/wicked`
```
ChaitanyaKulkarni28, line 137: nit:`/etc/sysconfig/network` instead of `/etc/sysconfig/wicked` need to adjust example path as well

drewhli: Done.

README.md
```diff
@@ -108,16 +108,68 @@ skew may result in `system time has changed` messages in VM logs.
...
 The guest agent uses network interface metadata to manage the network
-interfaces in the guest by performing the following tasks:
-
-*   Enabled all associated network interfaces on boot.
+interfaces by performing the following tasks:
+
+*   Enable all associated network interfaces on boot.
+    *   Detect the current active network manager service that is managing the
+        primary NIC.
+    *   Rollback and delete any files/configurations left behind by all other
```
dorileo, line 116: outline that we rollback guest-agent managed files - not user's.

README.md
```diff
@@ -108,16 +108,68 @@ skew may result in `system time has changed` messages in VM logs.
 #### Network
 
 The guest agent uses network interface metadata to manage the network
-interfaces in the guest by performing the following tasks:
-
-*   Enabled all associated network interfaces on boot.
+interfaces by performing the following tasks:
```
dorileo, line 111: please add a section mentioning MDS route (i.e. guest agent configurations guarantees that routes to MDS are only created for primary nic).

dorileo (changes requested):

README.md
drewhli: Done.

README.md
drewhli: Done.

PR #436, opened 2024-09-30
network: debian 12 rollback only if default netplan is ok

dorileo:
In debian 12 systems we must only rollback dangling/left over configurations if we can successfully restore the default netplan config. That guarantees that we have a system in a state where the primary NIC is functional.

It prevents us to end up with a system without a functional netplan *and* without a functional systemd-networkd configuration.

### DIFF
--- a/google_guest_agent/network/manager/manager.go
+++ b/google_guest_agent/network/manager/manager.go
@@ -190,22 +190,8 @@ func SetupInterfaces(ctx context.Context, config *cfg.Sections, mds *metadata.De
 		return fmt.Errorf("error detecting network manager service: %v", err)
 	}
 
-	// Remove only primary nics left over configs.
-	if !config.NetworkInterfaces.ManagePrimaryNIC {
-		primaryInterface := mds.Instance.NetworkInterfaces[0]
-		nic := &Interfaces{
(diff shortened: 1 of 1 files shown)


google_guest_agent/network/manager/manager.go
```diff
@@ -244,6 +230,32 @@ func SetupInterfaces(ctx context.Context, config *cfg.Sections, mds *metadata.De
 	return nil
 }
 
+// Remove only primary nics left over configs.
+func rollbackLeftoverConfigs(ctx context.Context, config *cfg.Sections, mds *metadata.Descriptor) error {
+	if config.NetworkInterfaces.ManagePrimaryNIC {
+		return nil
```
ChaitanyaKulkarni28, line 236: shouldn't this be restoring netplan configs regardless of we're managing primary nic or not to match current behavior?

dorileo: Yep, restoring netplan should happen regardless and nic rollback should be conditioned to both netplan restore and being managing primary nic. I'll adjust.

dorileo: done.

PR #438, opened 2024-10-04
Retry MDS PUT operation, reload netplan/networkctl only if configs are changed

ChaitanyaKulkarni28:
- Adds retry logic for MDS PUT operation to avoid flaky failures
- Consolidates logic of checking CLI existence across network managers in one function to avoid duplication
- To avoid unnecessary DHCP refresh, now agent does `networkctl reload` and `netplan apply` only if configs are really added/removed by agent
- Errors occurred during `Rollback` might not be real issues as we run rollback on all network managers, downgraded logs from error to warning to avoid concerns 

/cc @dorileo @drewhli

### DIFF
--- a/google_guest_agent/network/manager/manager.go
+++ b/google_guest_agent/network/manager/manager.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"time"
 
 	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/cfg"
 	"github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/osinfo"
@@ -202,7 +203,7 @@ func SetupInterfaces(ctx context.Context, config *cfg.Sections, mds *metadata.De
 
 		logger.Infof("Rolling back %s", svc.Name())
 		if err = svc.Rollback(ctx, nics); err != nil {
-			logger.Errorf("Failed to roll back config for %s: %v", svc.Name(), err)
+			logger.Warningf("Unable to roll back config for %s: %v", svc.Name(), err)
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


google_guest_agent/network/manager/manager.go
```diff
@@ -227,6 +228,9 @@ func SetupInterfaces(ctx context.Context, config *cfg.Sections, mds *metadata.De
 
 	logger.Infof("Finished setting up %s", activeService.manager.Name())
 
+	// Setup might not have finished when we log and collect this information. Adding this
+	// temporary sleep for debugging purposes to make sure we have up-to-date information.
+	time.Sleep(2 * time.Second)
```
dorileo, line 233: since we are only interested on logging and the only shared data is the context we could use a separate go routine and not block here, what do you think?

google_guest_agent/network/manager/manager.go
```diff
@@ -253,7 +257,7 @@ func rollbackLeftoverConfigs(ctx context.Context, config *cfg.Sections, mds *met
 
 	for _, svc := range knownNetworkManagers {
 		if err := svc.RollbackNics(ctx, nic); err != nil {
-			logger.Errorf("Failed to rollback primary nic (left over) config for %s: %v", svc.Name(), err)
+			logger.Warningf("Failed to rollback primary nic (left over) config for %s: %v", svc.Name(), err)
```
dorileo, line 260: Why are we changing the log level of all these loggings?

ChaitanyaKulkarni28: most of the errors are not real errors since we're doing rollback on all manager's (most of which are not applicable) so logging it at `Error` level might cause confusion and unnecessary concerns so I downgraded.

google_guest_agent/network/manager/manager.go
ChaitanyaKulkarni28: Done

google_guest_agent/network/manager/manager.go
dorileo: ok

PR #441, opened 2024-10-07
Fallback to default configs if MDS becomes inaccessible

ChaitanyaKulkarni28:
Blindly retrying constantly if MDS becomes inaccessible for some reason might not help, try falling back to default configs. If that doesn't work either do exit. Systemd would restart agent process which will attempt configuring the network again. 

/cc @dorileo @drewhli 

/hold

### DIFF
--- a/google_guest_agent/events/metadata/metadata.go
+++ b/google_guest_agent/events/metadata/metadata.go
@@ -19,7 +19,10 @@ import (
 	"context"
 	"net"
 	"net/url"
+	"strings"
+	"time"
 
+	network "github.com/GoogleCloudPlatform/guest-agent/google_guest_agent/network/manager"
 	"github.com/GoogleCloudPlatform/guest-agent/metadata"
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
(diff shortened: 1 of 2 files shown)


google_guest_agent/events/metadata/metadata.go
```diff
@@ -67,10 +69,28 @@ func (mp *Watcher) Run(ctx context.Context, evType string) (bool, interface{}, e
 			}
 			logger.Errorf("Error watching metadata: %s", err)
 			mp.failedPrevious = true
```
dorileo, line 71: after falling back do we still want to keep `failedPrevious = true` ?

dorileo: and for all cases?

ChaitanyaKulkarni28: ah yes, if fallback succeeds we should rest it, in case of failure it wouldn't matter as it would exit and restart.

ChaitanyaKulkarni28: updated `tryFallback()` to reset

PR #442, opened 2024-10-07
network: force reconfigure for changed interfaces

dorileo:
We want to force reconfigure of interfaces that had their configuration generated - including the primary interface.

### DIFF
--- a/google_guest_agent/network/manager/netplan_linux.go
+++ b/google_guest_agent/network/manager/netplan_linux.go
@@ -192,7 +192,7 @@ func (n *netplan) SetupEthernetInterface(ctx context.Context, config *cfg.Sectio
 	}
 
 	// Write the config files.
-	reload1, err := n.writeNetplanEthernetDropin(mtuMap, googleInterfaces, googleIpv6Interfaces)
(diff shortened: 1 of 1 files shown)


google_guest_agent/network/manager/netplan_linux.go
```diff
@@ -216,17 +216,32 @@ func (n *netplan) SetupEthernetInterface(ctx context.Context, config *cfg.Sectio
...
+		return nil
+	}
+
+	// Force reconfigure of changed interfaces and primary interface.
+	cmdArgs := []string{"reconfigure"}
+	for k := range interfaces {
+		cmdArgs = append(cmdArgs, k)
+	}
+
+	if err := run.Quiet(ctx, "networkctl", cmdArgs...); err != nil {
```
ChaitanyaKulkarni28, line 243: as per man page - `Note that this does not reload .netdev or .network corresponding to the specified interface. So, if you edit config files, it is necessary to call networkctl reload first to apply new settings.` Based on the documentation it doesn't seem like we can drop `reload` and just call `reconfigure`

dorileo: We are not dropping reload, but I agree that this will have little to no effect in fact.

PR #443, opened 2024-10-08
network/netplan: adjust link-local accordingly

dorileo:
Adjust the netplan's link-local configuration key accordingly with instance's enabled stack.

In general default values should be fine for most of netplan + networkd versions combinations, however, in debian-12 we've seen failures where the network interface will randomly not be available if the stacks are not properly adjusted.


google_guest_agent/network/manager/netplan_linux.go
```diff
@@ -342,23 +345,32 @@ func (n *netplan) writeNetplanEthernetDropin(mtuMap map[string]int, interfaces,
...
 				UseDomains: shouldUseDomains(i),
 			},
+			LinkLocal: []string{"ipv4"},
 		}
 
 		if mtu, found := mtuMap[iface]; found {
 			ne.MTU = &mtu
 		}
 
+		ne.DHCPv6 = &falseVal
```
ChaitanyaKulkarni28, line 363: as by default dhcpv6 is disabled only, do we need this section?

dorileo: indeed. adjusted.

PR #449, opened 2024-10-10
Skip linklocal settings on ubuntu 20.04

ChaitanyaKulkarni28:
Do not set LinkLocal on Ubuntu 20.04 it ends up adding route to MDS on secondary NIC which breaks all calls to MDS with error `dial tcp 169.254.169.254:80: i/o timeout`
/cc @dorileo @drewhli


ChaitanyaKulkarni28:
Closing this PR as we're reverting the change #448

PR #454, opened 2024-10-16
packaging: add plugin manager to rhel packaging

dorileo:
This change introduces the packaging bits to install both guest agent and plugin manager side-by-side for rhel systems.

### DIFF
--- a/packaging/google-guest-agent.spec
+++ b/packaging/google-guest-agent.spec
@@ -15,6 +15,11 @@
 # Don't build debuginfo packages.
 %define debug_package %{nil}
 
+# The only use for extra source is to build plugin manager.
+%if 0%{?has_extra_source}
+%define build_plugin_manager %{has_extra_source}
+%endif
+
 Name: google-guest-agent
 Epoch:   1
 Version: %{_version}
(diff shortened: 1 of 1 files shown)


dorileo:
> Packaging looks good but it seems to be missing `systemctl enable/start` on google-guest-agent-manager.service after install/upgrade

Yep, just pushed a new version with it (I was waiting the PR containing the .service file to be merged).

packaging/google-guest-agent.spec
```diff
@@ -36,14 +44,25 @@ Obsoletes: python-google-compute-engine, python3-google-compute-engine
...
 %build
 for bin in google_guest_agent google_metadata_script_runner gce_workload_cert_refresh; do
   pushd "$bin"
   GOPATH=%{_gopath} CGO_ENABLED=0 %{_go} build -ldflags="-s -w -X main.version=%{_version}" -mod=readonly
   popd
 done
+# Build side-by-side both new agent (plugin manager) and legacy agent.
+%if 0%{?build_plugin_manager}
+pushd %{name}-extra-%{version}/
+  make cmd/google_guest_agent/google_guest_agent
```
ChaitanyaKulkarni28, line 63: make accepts version now, can we pass `%{_version}` here as well?

dorileo: indeed. done.

PR #455, opened 2024-10-16
plugin manager: introduce the systemd service file

dorileo:
This will sit side-by-side with the guest agent, the requirements are a bit different to guest-agent's - hence the lack of PartOf.

### DIFF
--- a/google-guest-agent-manager.service
+++ b/google-guest-agent-manager.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Google Compute Engine Guest Agent Plugin Manager
+After=google-guest-agent.service
+Wants=google-guest-agent.service
+
+[Service]
+Type=notify
+ExecStart=/usr/bin/google_guest_agent_manager
+OOMScoreAdjust=-999
+Restart=always
+
+[Install]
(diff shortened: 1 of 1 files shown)


google-guest-agent-manager.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Google Compute Engine Guest Agent Plugin Manager
+# Start after network is online.
+#   Debian/Ubuntu 16.04: networking.service
+#   SLES/EL7: network.service (SLES via wicked.service)
+#   EL8: NetworkManager.service
+#   COS/Ubuntu 18.04+: systemd-networkd.service
+After=network-online.target syslog.service
+After=network.service networking.service NetworkManager.service systemd-networkd.service
```
ChaitanyaKulkarni28, line 9: can we have this run after guest-agent? That way Plugin Manager doesn't need to worry about Network not being ready case

dorileo: Yep, we can do it.

dorileo: done.

DanielRaapDev:
causes #481

PR #456, opened 2024-10-17
Add support for including agent manager in guest-agent package

ChaitanyaKulkarni28:
/cc @dorileo @drewhli

### DIFF
--- a/packaging/googet/google-compute-engine-windows.goospec
+++ b/packaging/googet/google-compute-engine-windows.goospec
@@ -7,6 +7,7 @@
   "description": "Google Compute Engine Windows agent",
   "source": "https://github.com/GoogleCloudPlatform/guest-agent/",
   "files": {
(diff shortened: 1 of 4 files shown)


packaging/googet/google-compute-engine-windows.goospec
```diff
@@ -36,7 +37,8 @@
   "sources": [{
       "include": [
         "GCEWindowsAgent.exe",
-	"GCEAuthorizedKeysCommand.exe",
+        "GCEWindowsAgentManager.exe",
+	      "GCEAuthorizedKeysCommand.exe",
```
dorileo, line 41: <tab> vs <space> ?

ChaitanyaKulkarni28: Done, its funny actually, IDE and Github is parsing differently, its looks fine only on either of those and not both,
On IDE it looks like this now :)
Screenshot 2024-10-17 at 4 15 32 PM

PR #457, opened 2024-10-17
Add plugin manager to debian build

drewhli:
/hold

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -22,12 +22,18 @@ override_dh_auto_install:
 	install -p -m 0644 instance_configs.cfg debian/google-guest-agent/usr/share/google-guest-agent
 	install -d debian/google-guest-agent/lib/systemd/system
 	install -p -m 0644 gce-workload-cert-refresh.timer debian/google-guest-agent/lib/systemd/system/
+	if [[ -d google-guest-agent ]]; then\
+		install -p -m 0644 google-guest-agent/cmd/google_guest_agent/google_guest_agent debian/google-guest-agent/usr/bin/google_guest_agent_manager;\
+	fi
 
 override_dh_golang:
 	# We don't use any packaged dependencies, so skip dh_golang step.
 
 override_dh_auto_build:
 	dh_auto_build -O--buildsystem=golang -- -ldflags="-s -w -X main.version=$(VERSION)" -mod=readonly
+	if [[ -d google-guest-agent ]]; then\
+		VERSION=$(VERSION) make -C google-guest-agent cmd/google_guest_agent/google_guest_agent;\
+	fi
 
 override_dh_installinit:
 	# We don't ship sysvinit files or need script changes for them.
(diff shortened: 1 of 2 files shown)


packaging/debian/rules
```diff
@@ -22,19 +22,28 @@ override_dh_auto_install:
 	install -p -m 0644 instance_configs.cfg debian/google-guest-agent/usr/share/google-guest-agent
 	install -d debian/google-guest-agent/lib/systemd/system
 	install -p -m 0644 gce-workload-cert-refresh.timer debian/google-guest-agent/lib/systemd/system/
+	if [[ -d google-guest-agent ]]; do\
```
ChaitanyaKulkarni28, line 25: is this the right path? I believe guest-agent and google-guest-agent repos would be side-by-side and the package build happens from root of guest-agent repo

drewhli: With how the startupscripts are set up for the build, this is unfortunately not the case. So the workaround right now is to check in the `google-guest-agent` repo within the `guest-agent` repo and go from there.

ChaitanyaKulkarni28: startup scripts are not yet configured to pull extra repo, can't we make it clone the way we want? It would be weird to have one repo within another

packaging/debian/rules
```diff
@@ -22,19 +22,28 @@ override_dh_auto_install:
...
+		install -p -m 0644 google-guest-agent/cmd/google_guest_agent/google_guest_agent debian/google-guest-agent/usr/bin/google_guest_agent_manager;\
+	fi
 
 override_dh_golang:
 	# We don't use any packaged dependencies, so skip dh_golang step.
 
 override_dh_auto_build:
 	dh_auto_build -O--buildsystem=golang -- -ldflags="-s -w -X main.version=$(VERSION)" -mod=readonly
+	if [[ -d google-guest-agent ]]; then\
+		make -C google-guest-agent cmd/google_guest_agent/google_guest_agent;\
```
dorileo, line 35: missing VERSION on the make call.

packaging/debian/rules
```diff
@@ -22,19 +22,28 @@ override_dh_auto_install:
 	install -p -m 0644 instance_configs.cfg debian/google-guest-agent/usr/share/google-guest-agent
 	install -d debian/google-guest-agent/lib/systemd/system
 	install -p -m 0644 gce-workload-cert-refresh.timer debian/google-guest-agent/lib/systemd/system/
+	if [[ -d google-guest-agent ]]; then\
+		install -p -m 0644 google-guest-agent/cmd/google_guest_agent/google_guest_agent debian/google-guest-agent/usr/bin/google_guest_agent_manager;\
```
dorileo, line 26: we are missing a install for the plugin manager's `.service` file.

dorileo: never mind it's being done in the systemd install override.

packaging/debian/rules
```diff
@@ -22,19 +22,28 @@ override_dh_auto_install:
...
+	fi
 
 override_dh_installinit:
 	# We don't ship sysvinit files or need script changes for them.
 
 override_dh_systemd_enable:
 	install -d debian/google-guest-agent/lib/systemd/system
 	install -p -m 0644 *.service debian/google-guest-agent/lib/systemd/system/
+	if [[ -d google-guest-agent ]]; then\
+		install -p -m 0644 google-guest-agent/build/configs/usr/lib/systemd/system/google-guest-agent.service debian/google-guest-agent/lib/systemd/system/google-guest-agent-manager.service;\
```
dorileo, line 45: the plugin manager `.service` file should be [google-guest-agent-manager.service from legacy guest agent code base](https://github.com/GoogleCloudPlatform/guest-agent/blob/main/google-guest-agent-manager.service).

packaging/debian/rules
```diff
@@ -22,19 +22,28 @@ override_dh_auto_install:
...
+	fi
 	install -d debian/google-guest-agent/lib/systemd/system-preset
 	install -p -m 0644 *.preset debian/google-guest-agent/lib/systemd/system-preset/
 	dh_systemd_enable google-guest-agent.service google-startup-scripts.service google-shutdown-scripts.service gce-workload-cert-refresh.timer
```
dorileo, line 49: we are missing `google-guest-agent-manager.service` here in the enable line.

dorileo, line 49: we are also missing a call to `dh_systemd_start` for google-guest-agent-manager in `override_dh_systemd_start`.

packaging/debian/rules
drewhli: Done.

packaging/debian/rules
drewhli: Resolving

packaging/debian/rules
drewhli: Done.

drewhli: Done.

packaging/debian/rules
drewhli: Done.

packaging/debian/rules
drewhli: The `.service` file should still be in the directory above since `google-guest-agent` is cloned into the `guest-agent` codebase, which becomes the root of the build staging directory.

dorileo: my point is that you are installing the wrong `.service` file, you are installing the one from new guest agent when you should be installing the one within legacy's code base.

drewhli: The legacy codebase's `.service` files should be covered by the line above,

PR #459, opened 2024-10-24
Add GOPATH to PATH in debian build

drewhli:
This should resolve the current issue where it cannot find `protoc-gen-go` and `protoc-gen-go-grpc` during the build process. This also addresses a `protoc` issue where it could not find or resolve the import dependencies of `proto` files.

### DIFF
--- a/packaging/debian/control
+++ b/packaging/debian/control
@@ -3,7 +3,7 @@ Maintainer: Google Cloud Team <gc-team@google.com>
 Section: misc
 Priority: optional
 Standards-Version: 3.9.8
-Build-Depends: debhelper (>= 9.20160709), dh-golang (>= 1.1), golang-go, protobuf-compiler
+Build-Depends: debhelper (>= 9.20160709), dh-golang (>= 1.1), golang-go
 
 Package: google-guest-agent
 Architecture: any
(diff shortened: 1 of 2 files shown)


packaging/debian/control
```diff
@@ -3,7 +3,7 @@ Maintainer: Google Cloud Team <gc-team@google.com>
 Section: misc
 Priority: optional
 Standards-Version: 3.9.8
-Build-Depends: debhelper (>= 9.20160709), dh-golang (>= 1.1), golang-go, protobuf-compiler
```
dorileo, line 6: why are you removing protobuf-compiler from build deps?

packaging/debian/rules
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/make -f
 
-export PATH := /tmp/go/bin:$(PATH)
+export PATH := /tmp/go/bin:/usr/share/gocode/bin:$(PATH)
```
dorileo, line 3: is this for you to use the protobuf pre loaded in the build image?

drewhli: Yea, so we can access `protoc-gen-go` and `protoc-gen-go-grpc`. Otherwise we see errors that say that those aren't executable or aren't in `PATH`.

packaging/debian/control
drewhli: We're installing `protoc` ourselves, so I didn't think we'd need it. However, with the most recent commit adding the downloaded `protoc` to path before the `apt` version of `protoc`, we could add it back.

PR #473, opened 2024-11-22
stable/network: fix cliExists() call pattern

dorileo:
In networkd and NetworkManager implementation we were wrongly assuming the existing of the cli based only the error return, with that we were ignoring one of the conditions where we don't return an error even though the cli doesn't exists.


dorileo:
sent it to main, it should be sent to stable.

PR #475, opened 2024-11-27
Update metadata script runner to honor cloud logging config flag

ChaitanyaKulkarni28:
`cloud_logging_enabled` config key was introduced to control enable/disable of Cloud Logging. This change updates metadata script runner to honor that flag.

/cc @dorileo @drewhli

### DIFF
--- a/google_metadata_script_runner/main.go
+++ b/google_metadata_script_runner/main.go
@@ -162,9 +162,7 @@ func downloadScript(ctx context.Context, path string, file *os.File) error {
 			return nil
 		}
 
-		if err != nil {
-			logger.Infof("Failed to download object [%s] from GCS bucket [%s], err: %+v", object, bucket, err)
-		}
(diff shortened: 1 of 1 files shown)


google_metadata_script_runner/main.go
```diff
@@ -162,9 +162,7 @@ func downloadScript(ctx context.Context, path string, file *os.File) error {
 			return nil
 		}
 
-		if err != nil {
-			logger.Infof("Failed to download object [%s] from GCS bucket [%s], err: %+v", object, bucket, err)
-		}
+		logger.Infof("Failed to download object [%s] from GCS bucket [%s], err: %+v", object, bucket, err)
```
dorileo, line 165: Can you add the context that we'll try unauth download to the log msg?

dorileo:
minor comment, other than that lgtm.

google_metadata_script_runner/main.go
ChaitanyaKulkarni28: There's already a log below this which state that -  `Trying unauthenticated download` (line 167)

PR #482, opened 2025-01-24
Do not fail network setup if there is an invalid/bad MAC

drewhli:

### DIFF
--- a/google_guest_agent/network/manager/common.go
+++ b/google_guest_agent/network/manager/common.go
@@ -86,9 +86,16 @@ func interfaceNames(nics []metadata.NetworkInterfaces) ([]string, error) {
 	for _, ni := range nics {
 		iface, err := GetInterfaceByMAC(ni.Mac)
 		if err != nil {
-			return nil, err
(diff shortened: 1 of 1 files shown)


google_guest_agent/network/manager/manager.go
```diff
@@ -194,9 +194,9 @@ func SetupInterfaces(ctx context.Context, config *cfg.Sections, mds *metadata.De
 		VlanInterfaces:     map[int]VlanInterface{},
 	}
 
-	interfaces, err := interfaceNames(nics.EthernetInterfaces)
-	if err != nil {
-		return fmt.Errorf("error getting interface names: %v", err)
+	interfaces := interfaceNames(nics.EthernetInterfaces)
+	if len(interfaces) == 0 {
```
dorileo, line 198: can we move this size check to the function and keep it returning an error? (so we don't have to duplicate it everywhere. I believe it's valid for all the cases - please check it).

drewhli: Done.

PR #486, opened 2025-01-30
vlan: don't index based on the vlan ID

dorileo:
The vlan ID may be repeated in a system as long as the vlan is set for different parent interfaces. This change makes sure to map based on the combination of parentID + Vlan (in the form of parentID-Vlan).

### DIFF
--- a/google_guest_agent/network/manager/common.go
+++ b/google_guest_agent/network/manager/common.go
@@ -108,18 +108,6 @@ func vlanInterfaceParentMap(nics map[int]metadata.VlanInterface, allEthernetInte
 	return vlans, nil
 }
 
-// vlanInterfaceListsIpv6 gets a list of VLAN IDs that support IPv6.
-func vlanInterfaceListsIpv6(nics map[int]VlanInterface) []int {
-	var googleIpv6Interfaces []int
-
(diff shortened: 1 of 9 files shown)


dorileo:
I'm adding a test case similar to the one identified by the integration tests.

robromano:
The vlan IDs only need to be unique per parent NIC.

google_guest_agent/network/manager/common.go
```diff
@@ -109,7 +109,7 @@ func vlanInterfaceParentMap(nics map[int]metadata.VlanInterface, allEthernetInte
 }
 
 // vlanInterfaceListsIpv6 gets a list of VLAN IDs that support IPv6.
-func vlanInterfaceListsIpv6(nics map[int]VlanInterface) []int {
+func vlanInterfaceListsIpv6(nics map[string]VlanInterface) []int {
```
ChaitanyaKulkarni28, line 112: shouldn't this return parent+vlan_id as list of ids otherwise we won't be able to differentiate between vlan nics?

dorileo: we should, however, we don't even need this function. I'm sending an updated version of this PR.

PR #489, opened 2025-01-31
Include interfaces in lists even if it has an invalid MAC.

drewhli:
Skip the interfaces when setting up interfaces and writing configs.

This also cleans up some unused VLAN functions.

/hold


google_guest_agent/network/manager/common.go
```diff
@@ -86,28 +83,16 @@ func interfaceNames(nics []metadata.NetworkInterfaces) ([]string, error) {
 	for _, ni := range nics {
 		iface, err := GetInterfaceByMAC(ni.Mac)
 		if err != nil {
-			return nil, err
+			if _, found := badMAC[ni.Mac]; !found {
```
ChaitanyaKulkarni28, line 86: we need to check all references of `interfaceNames()` , here in case of error we're writing empty string as interface name in the list (it won't have `invalid` string like in the other method) that might not work everywhere 

for e.g [here](https://github.com/GoogleCloudPlatform/guest-agent/blob/d13038f4f3dad8c0dc70c7c6fa5c5d77b67d1037/google_guest_agent/network/manager/network_manager_linux.go#L170) it gets the list and calls `writeNetworkManagerConfigs`, `strings.Contains(iface, "invalid")` won't be true and it will proceed to configure

google_guest_agent/network/manager/common.go
```diff
@@ -86,28 +83,16 @@ func interfaceNames(nics []metadata.NetworkInterfaces) ([]string, error) {
...
+				logger.Errorf("Error getting interface %s: %v", ni.Mac, err)
+				badMAC[ni.Mac] = iface
+			}
 		}
 		ifaces = append(ifaces, iface.Name)
```
ChaitanyaKulkarni28, line 91: we need to check all references of `interfaceNames()` , here in case of error we're writing empty string as interface name in the list (it won't have `invalid` string like in the other method) that might not work everywhere 

for e.g [here](https://github.com/GoogleCloudPlatform/guest-agent/blob/d13038f4f3dad8c0dc70c7c6fa5c5d77b67d1037/google_guest_agent/network/manager/network_manager_linux.go#L170) it gets the list and calls `writeNetworkManagerConfigs`, `strings.Contains(iface, "invalid")` won't be true and it will proceed to configure 

maybe instead of empty write `invalid` here as well?

drewhli: Done. Added `invalid`.

google_guest_agent/network/manager/common.go
```diff
@@ -85,27 +83,30 @@ func interfaceNames(nics []metadata.NetworkInterfaces) ([]string, error) {
...
-	for _, ni := range nics {
-		parentInterface, err := vlanParentInterface(allEthernetInterfaces, ni)
-		if err != nil {
-			return nil, fmt.Errorf("failed to determine vlan's parent interface: %+v", err)
-		}
-		vlans[ni.Vlan] = parentInterface
+// isInvalid checks if the provided interface is invalid. This is used to skip
+// writing configurations for interfaces that have been disabled or otherwise
+// made invalid. The `invalid` tag is added to ifaces whose MACs are invalid.
+// This logic is handled in the `interfaceListsIpv4Ipv6 function below.
```
dorileo, line 103: missing the closing `after interfaceListIpv4Ipv6. Please also expand the comment to what problem marking the interface as "invalid" solves, basically describe the bug we are fixing with it.

google_guest_agent/network/manager/common.go
```diff
@@ -116,17 +117,19 @@ func interfaceListsIpv4Ipv6(nics []metadata.NetworkInterfaces) ([]string, []stri
...
 		iface, err := GetInterfaceByMAC(ni.Mac)
+		ifaceName := iface.Name
 		if err != nil {
 			if _, found := badMAC[ni.Mac]; !found {
 				logger.Errorf("error getting interface: %s", err)
 				badMAC[ni.Mac] = iface
 			}
-			continue
+			// Mark the iface as invalid, and include its Mac.
+			ifaceName = fmt.Sprintf("invalid-%s", ni.Mac)
```
dorileo, line 127: describe why it's important to include this interface in the slice even though it's a invalid one.

PR #490, opened 2025-01-31
scripts: introduce a wrapper to locally build deb package

dorileo:
This change introduces a shell script to wrap the steps to locally build a deb package based on our debian packaging.

The user must provide VERSION and RELEASE environment variable to be passed in to deb builder.

### DIFF
--- a/build-deb.sh
+++ b/build-deb.sh
@@ -1,0 +1,47 @@
+#!/bin/env bash
+#
+# This script wraps the steps required to locally build a deb package
+# based on the in tree debian packaging (see invocation example bellow).
+#
+# After the build is finished the script will print out where the final
+# artifact was generated.
+#
+# Invocation:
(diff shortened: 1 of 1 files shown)


build-deb.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/env bash
+#
+# This script wraps the steps required to locally build a deb package
+# based on the in tree debian packaging (see invocation example bellow).
+#
+# After the build is finished the script will print out where the final
+# artifact was generated.
+#
+# Invocation:
+# VERSION=<version> RELEASE=<release> ./build-deb.sh
```
ChaitanyaKulkarni28, line 10: can we also document `COMMITURL`

dorileo: actually it's not needed. I'm dropping it.

PR #495, opened 2025-02-04
vlan: toggle vlan configuration in debian packaging

dorileo:
We are enabling vlan support by default for debian systems. Note we are not forcing overwriting the configuration for updated instances, user will need to manually toggle the configuration for long running instances if they need/want to enable it.

### DIFF
--- a/instance_configs.cfg
+++ b/instance_configs.cfg
@@ -36,6 +36,7 @@ dhclient_script = /sbin/google-dhclient-script
 dhcp_command = 
 ip_forwarding = true
 setup = true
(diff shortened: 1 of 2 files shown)


instance_configs.cfg
```diff
@@ -36,6 +36,7 @@ dhclient_script = /sbin/google-dhclient-script
 dhcp_command = 
 ip_forwarding = true
 setup = true
+# vlan_setup_enabled = false
```
ChaitanyaKulkarni28, line 39: why is this commented out? Default should anyways be `false`

PR #508, opened 2025-02-28
Add 'created-by' metadata, and pass it as option to logging library

drewhli:

This will need some `go.mod` and `go.sum` changes after the [respective change in the logging library](https://github.com/GoogleCloudPlatform/guest-logging-go/pull/40) has been merged.

/hold

### DIFF
--- a/gce_workload_cert_refresh/main.go
+++ b/gce_workload_cert_refresh/main.go
@@ -149,7 +149,10 @@ type outputOpts struct {
 
 func main() {
 	ctx := context.Background()
-	createdByBytes, _ := getMetadata(ctx, "/instance/attributes/created-by")
+	createdByBytes, err := getMetadata(ctx, "/instance/attributes/created-by")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error getting metadata attribute created-by: %v\n", err)
+	}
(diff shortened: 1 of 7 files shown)


dorileo:
please hold this until next week's release.

gce_workload_cert_refresh/main.go
```diff
@@ -149,12 +149,13 @@ type outputOpts struct {
 
 func main() {
 	ctx := context.Background()
-
+	createdByBytes, _ := getMetadata(ctx, "/instance/attributes/created-by")
```
ChaitanyaKulkarni28, line 152: should this log the error to stderr? Non-existing key is fine to ignore i guess but if mds connection had failed its probably good to at-least log it?

google_authorized_keys/main.go
```diff
@@ -143,9 +143,11 @@ func main() {
 	ctx := context.Background()
 	username := os.Args[1]
 
+	createdBy, _ := client.GetKey(ctx, "/instance/attributes/created-by", nil)
```
ChaitanyaKulkarni28, line 146: same as above

drewhli: Done.

PR #520, opened 2025-03-21
oslogin: Correctly handle newlines at the end of modified files

ajorg:
e38e673 oslogin: avoid adding extra empty line at the end of /etc/security/group.conf (#369)

had reverted:

b3376e8 end group.conf with newline (#64)

When really the problem that e38e673 was trying to solve was the handling of the terminal \n in filterGoogleLines()

### DIFF
--- a/google_guest_agent/oslogin.go
+++ b/google_guest_agent/oslogin.go
@@ -228,7 +228,7 @@ func cleanupDeprecatedLines(fpath string, directives []string) error {
 		return nil
 	}
 
-	err = os.WriteFile(fpath, []byte(strings.Join(updatedLines, "\n")), stat.Mode())
+	err = os.WriteFile(fpath, []byte(strings.Join(updatedLines, "\n")+"\n"), stat.Mode())
 	if err != nil {
 		return fmt.Errorf("failed to update deprecated configuration directives: %+v", err)
 	}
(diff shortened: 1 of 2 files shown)


google_guest_agent/oslogin.go
```diff
@@ -265,6 +265,14 @@ func filterGoogleLines(contents string) []string {
 			filtered = append(filtered, line)
 		}
 	}
+	// Unix text files should end with a final "\n"
+	// which strings.Split will account for with a terminal ""
+	// so we remove that here, to avoid adding more empty lines.
+	// But we can't assume that every file does end in a newline,
+	// so only remove it if it's empty as expected.
+	if len(filtered) > 0 && filtered[len(filtered)-1] == "" {
```
ericdand, line 273: Did you mean to compare with something other than the empty string here? I don't think `filtered[len(filtered)-1] == ""` will ever be true.

ericdand (changes requested):
Thanks for drafting this, Andrew!

google_guest_agent/oslogin.go
ajorg: A split on a string with empty lines will yield empty string elements. The `\n` is removed when splitting.

So yes, this is intentional. It catches the case where the file ends in a newline, because split includes the empty string _after_ that terminal newline.

ajorg: To clarify a bit more, that block is there to avoid accidentally deleting the last element when there is _not_ a terminal newline (in which case the final element might not be empty).

ericdand: Ah, nice, I'd missed that `filtered` was an array of strings, not simply a string. Sounds good.

PR #522, opened 2025-03-24
Re-enable disabled services if the core plugin was enabled

drewhli:
Most of the info regarding this change are in [this PR](https://github.com/GoogleCloudPlatform/guest-agent/pull/521).

### DIFF
--- a/packaging/google-guest-agent.spec
+++ b/packaging/google-guest-agent.spec
@@ -199,6 +199,15 @@ else
       systemctl restart google-guest-agent-manager.service >/dev/null 2>&1 || :
     %endif
   fi
+
+  # Re-enable the guest agent service if core plugin was enabled, since the
+  # service would have been disabled, and stay disabled post-upgrade.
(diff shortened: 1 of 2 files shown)


packaging/google-guest-agent.spec
```diff
@@ -193,6 +193,16 @@ else
     
   if [ -d /run/systemd/system ]; then
     systemctl daemon-reload >/dev/null 2>&1 || :
+
+    # Re-enable the guest agent service if core plugin was enabled, since the
+    # service would have been disabled, and stay disabled post-upgrade.
+    if [ ! -f "/usr/bin/google_guest_compat_manager" ]; then
```
ChaitanyaKulkarni28, line 199: can we move this check outside `/run/systemd/system` directory exists check?  We might run into same problem where this directory does not exist on mounted image

drewhli: Done.

PR #524, opened 2025-03-25
Wicked: Only reload interfaces for which configurations are written or changed.

drewhli:
`wicked ifreload all` fails if there are any lingering configuration files that apply to an interface that either does not exist or has been made invalid. The rollback logic in this case would fail to clean up these dangling files because no interfaces would match the name, thus leaving this hanging configuration file. This is a workaround for this case to avoid network setup failing so that the rest of the network setup like IP-forwarding route setup can complete.

/hold

### DIFF
--- a/google_guest_agent/network/manager/wicked_linux.go
+++ b/google_guest_agent/network/manager/wicked_linux.go
@@ -103,11 +103,13 @@ func (n *wicked) SetupEthernetInterface(ctx context.Context, cfg *cfg.Sections, 
 	}
 
 	// https://manpages.opensuse.org/Tumbleweed/wicked/wicked.8.en.html#ifreload_-_checks_whether_a_configuration_has_changed,_and_applies_accordingly.
-	// Apply any configuration changes on all interface. If unchanged ifreload
-	// does not touch specified interfaces.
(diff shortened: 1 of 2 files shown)


google_guest_agent/network/manager/wicked_linux.go
```diff
@@ -97,16 +97,19 @@ func (n *wicked) SetupEthernetInterface(ctx context.Context, cfg *cfg.Sections,
...
-	if err = n.writeEthernetConfigs(ifaces); err != nil {
+	changed, err := n.writeEthernetConfigs(ifaces)
+	if err != nil {
 		return fmt.Errorf("error writing wicked configurations: %v", err)
 	}
 
 	// https://manpages.opensuse.org/Tumbleweed/wicked/wicked.8.en.html#ifreload_-_checks_whether_a_configuration_has_changed,_and_applies_accordingly.
 	// Apply any configuration changes on all interface. If unchanged ifreload
 	// does not touch specified interfaces.
-	args := []string{"ifreload", "all"}
```
ChaitanyaKulkarni28, line 107: nit: can you update the above comment as well

drewhli: Done.

PR #532, opened 2025-04-25
delete the unused sshtrustedca_windows.go file

ericdand:
This is a small PR to clean up some unused code.

This file is not presently linked at all, and is ignored during builds. It seems to have been checked in with the intention of supporting cert-based SSH on Windows in the future. However, supporting SSH on Windows is not in OS Login's future plans at this time, especially since Microsoft themselves have been sunsetting support for SSH. Certificates, especially, are unlikely to be supported on Windows anytime in the near future.

A much larger change that significantly revamps the whole `sshtrustedca` package will follow this one.


ericdand:
Ah interesting -- Prow has a build script that does source this file. I guess I've got to clean that up first.

ericdand:
I've learned more about how conditional compilation works in the public Go compiler, and I realize that to fix this properly I ought to rearrange the code so that the entire OS Login module is omitted on Windows. This will be a much larger change than just deleting this funny little stub file here. 

I take back this PR for now; more work is needed.

PR #537, opened 2025-05-22
Enable debug logging for core plugin testing

ChaitanyaKulkarni28:
/cc @dorileo @drewhli

### DIFF
--- a/instance_configs.cfg
+++ b/instance_configs.cfg
@@ -40,3 +40,7 @@ vlan_setup_enabled = false
 
 [OSLogin]
 cert_authentication = true
+
+[Core]
+log_level = 4
+log_verbosity = 4


instance_configs.cfg
```diff
@@ -40,3 +40,7 @@ vlan_setup_enabled = false
 
 [OSLogin]
 cert_authentication = true
+
+[Core]
```
drewhli, line 44: Do we not also want/need the `client_debug_logging` flag? Or is that only for ACS debug logging?

ChaitanyaKulkarni28: yes, that's for acs (library) debug logging, messages are not coming from agent

drewhli (approved):

PR #542, opened 2025-06-04
Update packaging to install the new routes script to the respective dispatcher hooks.

drewhli:
/hold

This will fail until [this PR](https://github.com/GoogleCloudPlatform/google-guest-agent/pull/93) is merged.

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -32,6 +32,8 @@ override_dh_auto_install:
 		install -p -m 0755 google-guest-agent/cmd/ggactl/ggactl_plugin debian/google-guest-agent/usr/bin/ggactl_plugin;\
 		install -p -m 0755 google-guest-agent/cmd/core_plugin/core_plugin debian/google-guest-agent/usr/lib/google/guest_agent/core_plugin;\
 		install -p -m 0755 google-guest-agent/cmd/metadata_script_runner_compat/gce_compat_metadata_script_runner debian/google-guest-agent/usr/bin/gce_compat_metadata_script_runner;\
+		install -d debian/google-guest-agent/usr/lib/networkd-dispatcher/routable.d;\
+		install -p -m 0755 google-guest-agent/build/configs/google_guest_agent_routes_setup.sh debian/google-guest-agent/usr/lib/networkd-dispatcher/routable.d/;\
 	fi
 
 override_dh_golang:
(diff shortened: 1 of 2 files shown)


packaging/debian/rules
```diff
@@ -31,6 +31,12 @@ override_dh_auto_install:
 		install -p -m 0755 google-guest-agent/cmd/ggactl/ggactl_plugin_cleanup debian/google-guest-agent/usr/bin/ggactl_plugin_cleanup;\
 		install -p -m 0755 google-guest-agent/cmd/core_plugin/core_plugin debian/google-guest-agent/usr/lib/google/guest_agent/core_plugin;\
 		install -p -m 0755 google-guest-agent/cmd/metadata_script_runner_compat/gce_compat_metadata_script_runner debian/google-guest-agent/usr/bin/gce_compat_metadata_script_runner;\
+		install -d debian/google-guest-agent/usr/lib/networkd-dispatcher/routable.d;\
+		install -p -m 0755 google-guest-agent/build/configs/google_guest_agent_routes_setup.sh debian/google-guest-agent/usr/lib/networkd-dispatcher/routable.d/;\
+		install -d debian/google-guest-agent/etc/NetworkManager/dispatcher.d;\
```
ChaitanyaKulkarni28, line 36: Why do we need to install NM specific files in debian packaging? Same for SLES specific config script

packaging/google-guest-agent.spec
```diff
@@ -91,6 +91,15 @@ install -p -m 0755 %{name}-extra-%{version}/cmd/ggactl/ggactl_plugin_cleanup %{b
 install -p -m 0755 %{name}-extra-%{version}/cmd/google_guest_compat_manager/google_guest_compat_manager %{buildroot}%{_bindir}/google_guest_compat_manager
 install -p -m 0755 %{name}-extra-%{version}/cmd/core_plugin/core_plugin %{buildroot}%{_exec_prefix}/lib/google/guest_agent/core_plugin
 install -p -m 0755 %{name}-extra-%{version}/cmd/metadata_script_runner_compat/gce_compat_metadata_script_runner %{buildroot}%{_bindir}/gce_compat_metadata_script_runner
+
+# Dispatcher hook route setup.
+install -d /usr/lib/networkd-dispatcher/routable.d
+install -p -m 0755 %{name}-extra-%{version}/build/configs/google_guest_agent_routes_setup.sh %{buildroot}/usr/lib/networkd-dispatcher/routable.d/google_guest_agent_routes_setup.sh
```
ChaitanyaKulkarni28, line 97: Same as above

packaging/debian/rules
drewhli: I was mainly copying what `guest-configs` is doing, as we can see the NetworkManager scripts from `guest-configs` also living in an `/etc/NetworkManager/dispatcher.d/` directory on Debian instances.

We don't really expect people to move away from `systemd-networkd` to `NetworkManager`, but it may still be better to include it just in case.

drewhli: I've removed the extra things for now. I think we can determine if the use-case is valid to add the others later down the line.

packaging/google-guest-agent.spec
```diff
@@ -129,6 +133,10 @@ install -p -m 0644 90-%{name}.preset %{buildroot}%{_presetdir}/90-%{name}.preset
 %{_bindir}/google_guest_agent_manager
 %{_bindir}/ggactl_plugin
 %{_exec_prefix}/lib/google/guest_agent/core_plugin
+
+/usr/lib/networkd-dispatcher/routable.d/google_guest_agent_routes.sh
```
ChaitanyaKulkarni28, line 137: since we're not installing scripts for all network manager's we this section should be updated as well right?

PR #557, opened 2025-07-01
Build rollforward package to re-enable original agent and disable core plugin

ChaitanyaKulkarni28:
/cc @dorileo @drewhli


packaging/debian/postinst
```diff
@@ -13,6 +13,17 @@ if [ "$1" = "configure" ] ; then
...
+  # Disable and stop this service in favor of original guest-agent instead of core plugin.
+  systemctl disable 'google-guest-compat-manager.service' > /dev/null || true
+  systemctl stop 'google-guest-compat-manager.service' > /dev/null || true
+
+  # Stop core plugin.
+  ggactl_plugin coreplugin stop >/dev/null 2>&1 || :
+
+  # Enable the legacy agent again.
+  systemctl enable 'google-guest-agent.service' > /dev/null || true
+  systemctl start 'google-guest-agent.service' > /dev/null || true
```
dorileo, line 26: don't we need to restart plugin manager? is it enough doing `ggactl_plugin coreplugin stop` ? isn't there anything that would trigger plugin manager to reload core plugin?

packaging/debian/prerm
```diff
@@ -2,7 +2,7 @@
 
 if [ "$1" = "purge" -o "$1" = "remove" ] ; then
     systemctl stop google-guest-agent-manager >/dev/null 2>&1 || :
-    ggactl_plugin_cleanup all >/dev/null 2>&1 || :
+    ggactl_plugin dynamic-cleanup >/dev/null 2>&1 || :
```
dorileo, line 5: I assume this is unrelated, is that right?

packaging/googet/agent_install.ps1
```diff
@@ -101,11 +103,16 @@ try {
...
 
   if ($install_manager) {
-    Restart-Service $compat_manager -Verbose
-    Restart-Service $manager_name -Verbose
+    & sc.exe config $compat_manager start=disabled
+    Stop-Service $compat_manager -Verbose
+
+    Stop-Service $manager_name -Verbose
+    & $ggactl_exe coreplugin stop
+    Start-Service $manager_name -Verbose
```
dorileo, line 115: are we starting and enabling legacy agent somewhere else?

packaging/debian/postinst
ChaitanyaKulkarni28: restarting of plugin manager is taken care by package manager, [this](https://github.com/GoogleCloudPlatform/guest-agent/blob/8d064e90a2e7e044f0dfdefa024d8e20bb1468c5/packaging/debian/rules#L36) that we're installing will prevent its relaunch by plugin manager, but ensure previous process is stopped we call `ggactl_plugin coreplugin stop`

packaging/googet/agent_install.ps1
ChaitanyaKulkarni28: yes, above [line 106](https://github.com/GoogleCloudPlatform/guest-agent/blob/8d064e90a2e7e044f0dfdefa024d8e20bb1468c5/packaging/googet/agent_install.ps1#L106) is equivalent of `systemctl enable`

PR #572, opened 2025-09-10
Add extra debug logging around toggling OS Login

ericdand:
Add extra debug logging around toggling OS Login.

A coworker is having trouble with his guest agent not doing anything when he sets `enable-oslogin`; this ought to help us to tell what's going wrong.

This change additionally contains a number of fixes for build failures introduced by changes in recent versions of the Go language, mostly surrounding bad printf statements, plus also a fix to generate 1024-bit SSH keys (rather than only 256) in our tests.

### DIFF
--- a/google_guest_agent/system_windows.go
+++ b/google_guest_agent/system_windows.go
@@ -35,12 +35,12 @@ type (
 func init() {
 	key, _, err := registry.CreateKey(registry.LOCAL_MACHINE, regKeyBase, registry.WRITE)
 	if err != nil {
-		logger.Fatalf(err.Error())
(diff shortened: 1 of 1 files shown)


ericdand:
/retest

Looks like Prow didn't pick up my latest commit in that last run. Trying it again...

ericdand:
/retest 

That's better; the latest tests uncovered some new build failures, which I just fixed. Let's run those again now...

ericdand:
/lgtm

I think all I need now is approval from a teammate. :)

ChaitanyaKulkarni28 (approved):

PR #588, opened 2025-12-12
Disable certificates when security keys are enabled

bpl4vv:
When security keys metadata is enabled, certificates can be used to bypass the security key authentication mechanism. This change disables certificates if security keys are enabled.


dorileo:
Do we have an equivalent implementation for core plugin?

PR #607, opened 2026-04-21
Update dependencies

drewhli:
/cc @SaswatPadhi @shenpai35


go.mod
```diff
@@ -1,78 +1,78 @@
 module github.com/GoogleCloudPlatform/guest-agent
 
-go 1.25
+go 1.25.0
```
bkatyl: Why not 1.26? If it needs to say 1.25, why not 1.25.9?

drewhli: Changed to `1.26`.

go.mod
```diff
@@ -1,78 +1,78 @@
...
 
 replace github.com/GoogleCloudPlatform/guest-agent/metadata => ../metadata
 
 require (
-	cloud.google.com/go/storage v1.56.1
-	github.com/GoogleCloudPlatform/guest-logging-go v0.0.0-20250819203904-a74b641fa7fd
+	cloud.google.com/go/storage v1.62.1
+	github.com/GoogleCloudPlatform/guest-logging-go v0.0.0-20251007172312-d04f6358d858
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/go-ini/ini v1.67.0
```
shenpai35: does it also need to be gopkg.in/ini.v1 ?

drewhli: This is actually using that import I think, but I'll change it to be in line with the new guest agent.

drewhli: Done.

SaswatPadhi (approved):

PR #621, opened 2026-06-16
fix dependency between sshd and google-guest-agent-manager.service

akerekes:
Attempts to fix a race condition between sshd and the guest agent service where on rare occasions the agent is started after the sshd service and leads to failure in test [TestSSHServiceStartsAfterGuestAgent](https://github.com/GoogleCloudPlatform/cloud-image-tests/blob/64826317fd1e94c4c08a1f8e75435e8bd6692991/test_suites/ssh/ssh_service_test.go#L34)

### DIFF
--- a/google-guest-agent-manager.service
+++ b/google-guest-agent-manager.service
@@ -1,10 +1,9 @@
 [Unit]
 Description=Google Compute Engine Guest Agent Plugin Manager
-# Ensure we start before any SSH service or keygen process begins
-Before=sshd.service ssh.service sshd-keygen.target
-# Wait for network and group under syslog
+# Start before sshd in order to regenerate SSH host keys.
+Before=sshd.service ssh.service
 After=network-online.target syslog.service
-Wants=network-online.target
+After=network.service networking.service NetworkManager.service systemd-networkd.service
 
 [Service]
 Type=notify
@@ -14,4 +13,6 @@ Restart=always
 KillMode=process
 
 [Install]
+WantedBy=sshd.service
 WantedBy=multi-user.target
+WantedBy=network.service networking.service NetworkManager.service systemd-networkd.service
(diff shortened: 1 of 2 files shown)


google-guest-agent-manager.service
```diff
@@ -1,9 +1,11 @@
 [Unit]
 Description=Google Compute Engine Guest Agent Plugin Manager
-# Start before sshd in order to regenerate SSH host keys.
-Before=sshd.service
+# Force a hard structural requirement; sshd cannot boot without this service
+Requires=sshd-keygen.target
```
ChaitanyaKulkarni28: With hard requirements, does it still work as expected when the target does not exist? `sshd-keygen.target` is not present on debian: 

```
# systemctl status sshd-keygen.target
Unit sshd-keygen.target could not be found.
```

google-guest-agent-manager.service
```diff
@@ -1,9 +1,11 @@
...
+Before=sshd.service sshd-keygen.target
```
ChaitanyaKulkarni28: I wonder if we need `ssh.service` too? `sshd.service` is just an alias on some distros to `ssh.service`

shenpai35:
Can you add some context in the description so that it is easier to refer to later on?

google-guest-agent-manager.service
akerekes: Indeed this was causing problem on Debian, I've updated the config to avoid it.

google-guest-agent-manager.service
akerekes: Added `ssh.service` as well.

google-guest-agent-manager.service
```diff
@@ -13,6 +14,4 @@ Restart=always
 KillMode=process
 
 [Install]
-WantedBy=sshd.service
 WantedBy=multi-user.target
-WantedBy=network.service networking.service NetworkManager.service systemd-networkd.service
```
ChaitanyaKulkarni28, line 18: do we need to remove this `WantedBy` on networking targets, I don't think its related to ssh ordering issue, is it?

akerekes: reverted this change

ChaitanyaKulkarni28 (approved):

PR #628, opened 2026-08-05
Fix SUSE Linux Micro 6+

rjschwei:
SUSE Linux Micro 6+ is based on the SLES 16 code base and we need to account for the same semantics as in SLES 16. Recognize SL Micro and handle it appropriately. Thanks to @Lidong Zhong from the SUSE L3 team.


google_guest_agent/oslogin.go
```diff
@@ -140,8 +140,9 @@ func (o *osloginMgr) Disabled(ctx context.Context) (bool, error) {
 }
 
 func setupSles16OSLoginDirs() error {
-	isSles16 := strings.Contains(osInfo.OS, "sles") || strings.Contains(osInfo.OS, "opensuse")
-	if !isSles16 || osInfo.Version.Major != 16 {
+    isSles16 := strings.Contains(osInfo.OS, "sles") || strings.Contains(osInfo.OS, "opensuse")
+    isSLMicro := := strings.Contains(osInfo.OS, "sl-micro")
```
ChaitanyaKulkarni28: Thanks for the PR @rjschwei 

There seems to build failure due to duplicate `:=`.

Also, while we're modifying this do you mind changing this to something like this for readabilit

```
isSles := strings.Contains(osInfo.OS, "sles") || strings.Contains(osInfo.OS, "opensuse") || strings.Contains(osInfo.OS, "sl-micro")

if !isSles || osInfo.Version.Major != 16 {
    return nil
}
```

Based on [this](https://documentation.suse.com/releasenotes/sle-micro/html/releasenotes_sle-micro_6.2/index.html#jsc-PED-12483) major version from `VERSION_ID` for `sl-micro` is also `16`

rjschwei: @ChaitanyaKulkarni28 I fixed the fat finger issue with the double ":=" The table you reference is not relevant to this. The table was created after Micro 6.2 was released and os-release will not change retroactively. Meaning the versions for micro 6+ are 6.0, 6.1 and 6.2. There will not be a Micro 6.3. Going forward starting with SLES 16.1 there will be an "immutable" mode that replaces the Micro product.

rjschwei:
@ChaitanyaKulkarni28 sorry need help with the tests, could you please take a look. I added tests to check Micro. Thanks

ChaitanyaKulkarni28 (approved):

PR #629, opened 2026-08-12
Remove KillMode from google-guest-agent-manager

SaswatPadhi:
KillMode=process only killed the manager, but left the children (individual plugin processes) running.


shenpai35 (approved):
From discussion, the original plan was for extensions to run as separate child processes that keep running even when the parent (the plugin manager) is killed. And now it doesn't make sense for any extensions to be running if the extension manager itself isn't.

PR #635, opened 2026-08-28
Wait for new account to become visible before adding to groups

pdostal:
`createGoogleUser` runs `useradd` and then immediately calls `gpasswd -a`
for each configured group. On systems where the account database is
fronted by an NSS cache (e.g. `nscd`) that happens to be restarting at
that exact moment, the group-membership lookup can transiently report
the freshly created user as nonexistent even though `useradd` succeeded.

When this races with the very first metadata poll on boot, the account
is left without its `google-sudoers` membership and without its SSH key
ever written (`updateAuthorizedKeysFile` is skipped because the caller
treats the missing `gUsers` entry as already handled), so the instance
is unreachable over SSH until an unrelated metadata change happens to
retrigger account provisioning.

This was observed causing sporadic `Permission denied (publickey)`
failures on boot, reproducible via the guest-agent's own log:

    google_guest_agent[...]: Creating user testuser.
    google_guest_agent[...]: ERROR non_windows_accounts.go:149 Error creating user: gpasswd: user 'testuser' does not exist.

right as `nscd` restarted (triggered by an unrelated network reconfig)
at the same moment. The instance boots fine, network comes up, sshd is
running — the only thing missing is the account being visible in time
for `gpasswd`.

## Fix

Add a short bounded poll, `waitForUserVisible`, after `createUser`
succeeds, using the existing `userExists()` to confirm the account is
actually visible before proceeding to group membership. Indirection
through `userExistsFunc` keeps this covered by unit tests without
touching the real user database.

## Testing

- Added `non_windows_accounts_test.go` covering: immediate success,
  eventual success (mirrors the observed race), permanent failure, and
  context cancellation.
- `go build ./...`, `go vet ./...`, `gofmt -l .` clean.
- Full existing `go test ./...` suite still passes.


pdostal:
Hello @drewhli, @SaswatPadhi, @ChaitanyaKulkarni28, can you please help me to process this?

drewhli (approved):

pdostal:
@drewhli the `prow/presubmit/gobuild` is now green - I've just rebased from the main branch.

shenpai35 (approved):

